PR #24, opened 2019-02-08
Added package table of contents.

justinformentin:
I added a table of contents to more easily jump to specific packages in the README, and slightly reorganized a couple of things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,14 @@
 <br/>
 <br/>
 
-[![Build Status](https://travis-ci.org/imbhargav5/rooks.svg?branch=master)](https://travis-ci.org/imbhargav5/rooks)
-
 <br/> 
 
 Collection of regularly used custom hooks as utils for React
 
 ## [Website](https://react-hooks.org)
+### Rooks website built with next.js.
+
+<br/>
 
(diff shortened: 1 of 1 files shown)


imbhargav5:
This is an excellent start. Will review it once I wake up!

imbhargav5:
@justinformentin Since the readme is autogenerated right now, it will get overridden by the autogeneration logic for the next time. Can we add the package table of contents to this file instead? 
https://github.com/imbhargav5/rooks/blob/master/helpers/update_readme.js
And then, `node helpers/update_readme` from the root of the project. It will generate a new Readme with the changes. 

Also, to move `rooks` package to the top, we will need to update the same `update_readme.js` file to make sure that `rooks` package has more importance than the rest.

justinformentin:
Makes sense!

imbhargav5:
Hey @justinformentin I made some changes to the readme to ensure that we can just update the Readme file instead of update_readme script. Do you want to make any changes to the current readme?

justinformentin:
Nope, looks good!

imbhargav5:
I feel bad that this PR has gotten stale. 😓 Would you like to try another issue @justinformentin ?

PR #32, opened 2019-02-11
Change keys of Options to optional.

yesmeck:
All these keys should be optional.


imbhargav5:
@yesmeck Awesome, Thanks!

PR #35, opened 2019-02-12
Fix Options type for visibility-sensor

fintara:
* make Options partial (I think it's clearer than having all keys ending with `?`)
* fix types according to documentation
* small formatting issue in readme


imbhargav5:
Ah! Looks good to me!

imbhargav5:
Thank you @fintara .

PR #47, opened 2019-02-14
Use `incrementBy` instead of `setCounter`

deadcoder0904:
Not needed but why not as it matches with the rest of the implementation


imbhargav5:
LGTM!

Issue #36, opened 2019-02-12
Ideas for more hooks!


imbhargav5:
Hi Community!,

What kind of hooks do you want to see? Let's pitch some ideas here and take them up!

Dianoga:
I'd like to see a `useDebounce` hook like this one https://github.com/streamich/react-use/blob/master/src/useDebounce.ts. 

I was going to submit a PR, but the mono-repo stuff is blocking me and I don't have time to figure it out right now. The `react-use` library is under `unlicense` so you should be able to bring it in directly if you decide you like the idea.

imbhargav5:
Definitely. Can you explain what is making it difficult for you to contribute? I will document it for everyone.

Dianoga:
I was able to create the hook via `yarn create`, but I don't know what needs to be done to get the new hook into the rooks package. Also, normally to test npm packages where I want to use them, I can just symlink the package into `node_modules` but in this case it wasn't obvious what should be linked.

simbathesailor:
Maybe a hook which listens for multiple key presses and does something when certain keys are pressed at a time.
I created a hook for listening to multiple keypress shortcuts on a page or any dom element where key press makes sense in this library. https://github.com/simbathesailor/react-key-collector .  I think this repo is suitable place for this hook
Maybe I can raise a pr for it after having discussions ?.

imbhargav5:
Definitely go ahead! Perhaps we can make a useKeys hook 🤔 ?

simbathesailor:
Cool , i will raise a pr asap and we will take it from there.

simbathesailor:
Hi @imbhargav5 
I was working on some react project, and somehow found a need for getting geolocation.
Can we have something called useGeolocation(). Most of the time its like a utility that get called when someone clicks on use my location text. But do we have any usecases where we need to have geolocation info(lat and lng ) on component mount or updates or something? What are your thoughts?

simbathesailor:
@imbhargav5 
How about of hook which provides the history of any variable.(useHistory) It can be used to implement undo functionality for lot many cases. 
Similar to https://github.com/omnidan/redux-undo.
 
 We have usePrevious but that  work for just previous value. 
Your thoughts ?

imbhargav5:
I think implementing useUndo might be a touch too much for this project since useUndo needs to track every state or props change in the app which we can't control. Redux surrounds the app so it would make sense to be part of the react-redux project.

simbathesailor:
[...]

PR #48, opened 2019-02-14
Add debounce hook

Dianoga:
This is probably incomplete, but the hook itself should work. Submitting based on request in https://github.com/imbhargav5/rooks/issues/36

### DIFF
--- a/packages/debounce/.babelrc
+++ b/packages/debounce/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env", "@babel/preset-react", "babel-preset-minify"]
+}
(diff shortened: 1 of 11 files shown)


imbhargav5:
Awesome! So from what I see in the PR, there are a few things we should fix. 

- [ ] Test cases
- [ ] JS Doc style comments (Please refer other hooks in the project for reference)
- [ ] If the arguments of the custom hook change, then it should clear the timeouts and create new ones. 
- [ ] Readme with example (We can do this, once the ones at the top are done.)

imbhargav5:
Let me know if there is something I can help you with to make things easy in the PR.

imbhargav5:
Added in commit c077f36d01fc9e1c1654076952fa46416df619e5. 😄 
Thank you!

PR #51, opened 2019-02-21
for useOutsideClick add bug fix for IOS that occurs only on the real …

theskillwithin:
…device

Even though this appears to work even in xcode ios emulator.  on a REAL ios device this fix is needed for it to function.

```
  if ('ontouchstart' in document.documentElement) {
    document.body.style.cursor = 'pointer'
  }
```

### DIFF
--- a/packages/outside-click/src/index.js
+++ b/packages/outside-click/src/index.js
@@ -1,6 +1,9 @@
 import { useLayoutEffect } from "react";
 
 function useOutsideClick(ref, handler, when = true) {
+  if ('ontouchstart' in document.documentElement) {
+    document.body.style.cursor = 'pointer'
+  }
   function handle(e) {
     if (ref && ref.current && !ref.current.contains(e.target)) {
       handler(e);


imbhargav5:
Awesome. Let's take this a step further and add the handler for touch events too? Right now it's just click. So essentially, we can add the event listener for the `touchstart `. What say?

theskillwithin:
@imbhargav5 the if ontouchstart just detects if its a mobile device.  the `document.body.style.cursor = 'pointer'` is the real fix

imbhargav5:
Oh. Will that be enough? I was thinking we might need `document.addEventListener("ontouchstart",cb)`

theskillwithin:
@imbhargav5 yes that will be enough.  

`document.addEventListener("ontouchstart",cb)` would be another way to fix this issue, but i find this way cleaner.

imbhargav5:
Wow. Awesome @theskillwithin . I am going to test it and merge it. Thank you 🎉

PR #59, opened 2019-04-01
correct typings for useToggle

elmarburke:
The typings didn't matched the docs, all parameter were required. This PR also adds the ability to use generic values.


imbhargav5:
Hi @elmarburke Thank you for the PR! I was a sick and couldn't reply sooner. Sorry about that.

So, we changed the useToggle hook to now use a reducer instead. This was from another PR and we added types there. I ran into a peculiar issue with generating types and I can use your inputs on that. I am not as experienced with typescript haha.

PR #68, opened 2019-06-09
Feat/use keys

imbhargav5:
- [x] Use keys hook
- [ ] Tests

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
     "packages/*"
   ],
   "useWorkspaces": true,
-  "version": "3.0.3-alpha.1"
+  "version": "3.1.0"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
     "packages/*"
   ],
   "useWorkspaces": true,
-  "version": "3.1.0"
+  "version": "3.1.1-alpha.0"
 }
--- a/packages/boundingclientrect/package.json
+++ b/packages/boundingclientrect/package.json
@@ -29,7 +29,7 @@
     "rooks",
     "react-hooks.org"
   ],
-  "version": "3.1.0",
+  "version": "3.1.1-alpha.0",
   "_id": "@rooks/use-boundingclientrect@",
   "publishConfig": {
     "access": "public"
--- a/packages/boundingclientrect/package.json
+++ b/packages/boundingclientrect/package.json
@@ -29,7 +29,7 @@
     "rooks",
     "react-hooks.org"
   ],
-  "version": "3.0.1",
+  "version": "3.1.0",
   "_id": "@rooks/use-boundingclientrect@",
   "publishConfig": {
     "access": "public"
--- a/packages/counter/package.json
+++ b/packages/counter/package.json
@@ -29,7 +29,7 @@
     "react-hooks.org",
     "counter hook"
   ],
-  "version": "3.0.3-alpha.1",
+  "version": "3.1.0",
   "publishConfig": {
     "access": "public"
   },
--- a/packages/counter/package.json
+++ b/packages/counter/package.json
@@ -29,7 +29,7 @@
     "react-hooks.org",
     "counter hook"
   ],
-  "version": "3.1.0",
+  "version": "3.1.1-alpha.0",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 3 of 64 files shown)


imbhargav5:
@simbathesailor Can we try adding tests to this?
Also, does this hook work if keys are pressed and held?

simbathesailor:
@imbhargav5  Yes, I will write test cases for these !!, And yes this hooks will work even if the keys are pressed and held. But will only trigger once when the condition is matched. You can see the behaviour in this example: https://codesandbox.io/s/p9q7q1l8y7.

I made some type fixes for these.  From where I can raise the pr  . can i use the same branch used previously ?.

imbhargav5:
Hey @simbathesailor I added you as collaborator. Can you work on this PR branch?

simbathesailor:
@imbhargav5 Thanks for adding me as a collaborator. I will work on this branch feat/use-keys for writing test cases by this weekend sometime.

simbathesailor:
Hey @imbhargav5 ,quite busy days. I will try to put in some efforts this weekend on writing test cases. I have not used react-testing-library before and even jest. But I think I would be able to write few test cases. Being positive :)

simbathesailor:
Hey @imbhargav5 , I have been trying to write test cases for useKeys, but failing to simulate the multiple keypresses. I have asked at spectrum also for the query I have.
I have also added the test file in code sandbox. https://codesandbox.io/s/p9q7q1l8y7.
I am not able to figure out from documentation how to simulate this, Any thing which i am doing wrong. Can you or someone  point it out ?  

https://spectrum.chat/testing-library/help-react/react-testing-library-custom-hook-dom-events~5b5e29d3-cd1b-4a3f-bb82-9997a8cd30ee
Thanks

simbathesailor:
I was able to solve the issue with testing of useKeys but on codesandbox. Same test case is running on sandbox perfectly, but not with the local repo https://codesandbox.io/s/p9q7q1l8y7. Any ideas?
I will also push the updated code .

imbhargav5:
@simbathesailor Awesome work. You can run tests for all packages by running `yarn test` at the root of the project or you can jump into each individual package and run `yarn test` there.

simbathesailor:
@imbhargav5 . I was able to write test cases but the test fails to start for me. I was able run test cases on codesandbox, but not in the rooks repo. I was already doing what you suggested. Here is the snapshot of the error . Do you have any idea why this is happening. should i setiup ts-jest for running test cases ?

Screenshot 2019-06-30 at 7 36 04 PM

simbathesailor:
@imbhargav5 Figured out. Needed to add ts-jest with preset "ts-jest/presets/js-with-babel" at root level. Then every thing works. Let me know if any thing better I can do. Here is the snapshot 
Screenshot 2019-07-01 at 10 10 40 PM

I hope you may like it :)

https://kulshekhar.github.io/ts-jest/user/config/#basic-usage

simbathesailor:
@imbhargav5 Added the testcases and story book entry also. Please have a look. I faced certain issues while rebasing with dev, hence not did that. The current commit doesnot have any conflict with the dev branch. There was one mistake for readme import instead of README import due to which story book was not spinning up properly. Have a look and please provide feedback.
Thanks.

imbhargav5:
@simbathesailor Everything looks good. Just one thing. `TPressedKeyMapping` should be internal to the hook. Otherwise all invocations of this hook across components will be able to access this object. I think we should avoid that. What do you say?

simbathesailor:
> @simbathesailor Everything looks good. Just one thing. `TPressedKeyMapping` should be internal to the hook. Otherwise all invocations of this hook across components will be able to access this object. I think we should avoid that. What do you say?

`type TPressedKeyMapping = {
  [key: string]: boolean;
};`
Sorry, I am a bit newbie with typescript. Will writing the types like this, will expose it to the consumer of the library. How's that happening. Is it the ts compiler generating picking up these types and exporting it. How can we avoid it? I will also read about it 

Thanks

imbhargav5:
@simbathesailor Never mind. Let me see if I can put these changes in.

imbhargav5:
@simbathesailor Hey I added continuous mode. :D It's amazing!

simbathesailor:
yes, this is amazing 
I will hopefully work on finalizing the intersectionObserver API next week. 
Cheers finally my pr is getting some shape.

PR #72, opened 2019-07-29
refactor: store object and arrays

harshzalavadiya:
- This PR will allow user to store Object, Arrays, Number etc. as well
- Basically this hook is already works as wrapper so we can wrap `JSON.parse()` and `JSON.stringify()` for ease of developers.

### DIFF
--- a/packages/localstorage/src/useLocalstorage.ts
+++ b/packages/localstorage/src/useLocalstorage.ts
@@ -1,8 +1,8 @@
 import { useState, useEffect } from "react";
 
 interface LocalStorageHandler {
-  value: string | null;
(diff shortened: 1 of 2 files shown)


imbhargav5:
@harshzalavadiya Thank you for contributing! Means a lot. This looks good. Can we also do something similar for useSessionStorage?

imbhargav5:
Actually, don't worry about it. I will do the sessionstorage part. I am going to merge this.

harshzalavadiya:
Hi @imbhargav5,
Didn't saw message but thanks for merging PR :smile:
If I get time then I'll submit PR for sessionstorage as well

Issue #86, opened 2019-08-19
Use `useEffect` instead of `useLayoutEffect` for outside-click hook


imbhargav5:
**Describe the bug**
useLayoutEffect is unnecessary for outside-click hook. And creates needless warnings on the server side.

PR #87, opened 2019-08-20
Feature/outside click hook update

maciekgrzybek:
Pull request for #86. I've also added some instructions into contributing file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,14 @@
 ### Creating a new hook
 
+#### New hook
 From the root of the project run `yarn run create`. This should start the process to create a new hook and set up the package.
+
+#### Current hooks fixes / updates
+Make sure you have Yarn package manager installed globally on your machine.
+
+From the root of the project:
+
+- run `yarn install`
+- run `yarn run build`
+- run `yarn run dev`
+
--- a/packages/outside-click/src/useOutsideClick.ts
+++ b/packages/outside-click/src/useOutsideClick.ts
@@ -1,4 +1,4 @@
-import { useLayoutEffect, MutableRefObject, useRef, useEffect } from "react";
+import { useEffect, MutableRefObject, useRef } from "react";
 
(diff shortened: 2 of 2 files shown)


imbhargav5:
@maciekgrzybek Everything looks good. Except the useEffect which has savedCallback.

maciekgrzybek:
Hmm, but the only time when `savedCallback.current` should be updated, would be the change of `handler` right? So it could live all in one `useEffect`. Don't take that as an attack or something :) just want to understand what's going on :)

maciekgrzybek:
Also, wouldn't `useCallback` be a good choice in this example? - https://reactjs.org/docs/hooks-reference.html#usecallback

maciekgrzybek:
I did some testing, both with single useEffect and with useCallback:
```
  const memoizedCallback = useCallback(
    (e) => {
      handler(e);
    },
    [],
  );
```
and it works fine in both cases. It always passes the correct event object (correct target). But maybe I'm missing something? Let me know, thanks :)

imbhargav5:
@maciekgrzybek I remember thinking about useCallback but I had to go for a ref for a reason I can't seem to remember. Let me see if I can reproduce the issue.

imbhargav5:
So useCallback will result in an unnecessary rerender which we don't need. We only need to update the function and not worry about rerender. And refs can update without rerender. So we should stick to `ref` when it comes to the savedCallback.
As far as the `savedCallback` in the same useEffect is concerned, you are right, I think putting in the same useEffect will be fine. Also, can you do one more change? I think `handle` can be within a useCallback.

maciekgrzybek:
Hey, @imbhargav5 I've just pushed the changes, hope that was what you were thinking about :)

imbhargav5:
Yep. Looks good. ❤️  It's a nuance but can you move memoized callback definition right before the useEffect and I will merge this.

maciekgrzybek:
Done :)

PR #90, opened 2019-08-21
using array instead of object so variables can be renamed easily

harshzalavadiya:
## Why this change?
- Using array over object for destructring allows user to easily use flexible variable names
- Official react also uses the same approach for hooks like [`useEffect`](https://reactjs.org/docs/hooks-effect.html).

```jsx
// Previously
const {value, set, remove} = useLocalstorage("my-value", 0);

// With array flexible variable naming
const [myValue, setMyValue, removeMyValue] = useLocalstorage("my-value", 0);
```

### DIFF
--- a/packages/localstorage/README.md
+++ b/packages/localstorage/README.md
@@ -26,10 +26,10 @@ import useLocalstorage from "@rooks/use-localstorage";
 
 ```jsx
 function Demo() {
-  const { value, set, remove } = useLocalstorage("my-value", 0);
+  const [value, set, remove] = useLocalstorage("my-value", 0);
   return (
     <p>
-      Value is {value}{" "}
+      Value is {value}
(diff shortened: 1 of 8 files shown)


imbhargav5:
Hey @harshzalavadiya awesome initiative. We should definitely support arrays but can we also support the original syntax? `useInterval` already does that.

harshzalavadiya:
Sounds good,
That way it can be backward compatible
Let me work on that 😄

imbhargav5:
@harshzalavadiya Awesome looks very good at first glance. Will take a closer look in a bit.

imbhargav5:
@harshzalavadiya Can you add storybook entries with the new array type usage for localstorage and session storage?

harshzalavadiya:
@imbhargav5 
Agreed first I thought of adding example but due to storybook-source not enabled there was no difference visible on storybook so I have enabled that as well now

PR #94, opened 2019-09-01
Feature/use throttle hook

maciekgrzybek:


imbhargav5:
Looks solid !! Going to merge it after I review it again.

maciekgrzybek:
nice one :)

imbhargav5:
I think we might need to change the hook's logic. Because setTimeout is hanging in there without a way to clear it.

maciekgrzybek:
What if I put logic inside of useEffect, so I could clear it there?

imbhargav5:
@maciekgrzybek Yes that's right. We need to figure out a way to put in the logic in the useEffect and clear the timeout in the cleanup part of the useEffect.

imbhargav5:
Oops accidentally closed it.

maciekgrzybek:
Hey, @imbhargav5 I've made the change, should be all right now :)

imbhargav5:
Hmmm, I think we need to work on the logic a bit more. let me come up with something and explanation for it on Monday.

maciekgrzybek:
Ok, what's wrong?

imbhargav5:
Let's track this in a different branch `feat/use-throttle`

PR #100, opened 2019-09-03
useGeolocationhook

simbathesailor:
Hi,
This is the implementation of useGeolocation hook. Let's review it and move this forward.

1. Add useGeolocationhook

2. Add test cases.

3. Add storybook example.

### DIFF
--- a/packages/geolocation/.babelrc
+++ b/packages/geolocation/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env", "@babel/preset-react", "babel-preset-minify"]
+}
--- a/packages/geolocation/.eslintrc
+++ b/packages/geolocation/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": ["rooks"]
+}
--- a/packages/geolocation/.npmignore
+++ b/packages/geolocation/.npmignore
@@ -1,0 +1,3 @@
+src
+test
+*.log
--- a/packages/geolocation/README.md
+++ b/packages/geolocation/README.md
@@ -15,16 +15,60 @@ npm install --save @rooks/use-geolocation
 ### Importing the hook
 
 ```javascript
-import useGeolocation from "@rooks/use-geolocation"
+import useGeolocation from "@rooks/use-geolocation";
 ```
 
 ### Usage
 
-```jsx
-function Demo() {
-  useGeolocation();
-  return null
-}
+### 1. Getting geolocation in a component
 
-render(<Demo/>)
+```jsx
+function App() {
+  const geoObj = useGeolocation();
+
+  return (
+    <div
+      style={{
+        display: "flex",
+        alignItems: "center",
+        flexDirection: "column"
+      }}
+    >
+      <p>{geoObj && JSON.stringify(geoObj)}</p>
+    </div>
+  );
+}
+render(<App />);
+```
+
+### 1. Getting geolocation in a component on some condition
+
+```jsx
+function App() {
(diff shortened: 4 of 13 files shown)


imbhargav5:
Looks solid. Let's keep `useGeolocation` simple. It's not used as much as `useOnOutsideClick` , so let's get rid of `when` option there. If people want to use it, they can just put it inside a component and render that component conditionally.

simbathesailor:
> Looks solid. Let's keep `useGeolocation` simple. It's not used as much as `useOnOutsideClick` , so let's get rid of `when` option there. If people want to use it, they can just put it inside a component and render that component conditionally.

Hi @imbhargav5 ,
Instead of removing the support of when, can we make it true by default without sending the when parameter. I mean instead of adding a component and conditionally rendering it, the consumer can pass when as false in that specific case. by making when as true by default when not sent in param, will make it the same API that you suggested. What's your take on this ?

simbathesailor:
> /rebase

so rebasing with dev, gave me some whitespace errors no conflicts errors as such. I will do it again and will push it.

imbhargav5:
As far as this hook is concerned and some other hooks which are top lovel and not used too often within components the when option just adds more complexity to the user to remember the API. We should avoid it as much as possible since it only takes a line more for the user to conditional render Irma component.

simbathesailor:
> As far as this hook is concerned and some other hooks which are top lovel and not used too often within components the when option just adds more complexity to the user to remember the API. We should avoid it as much as possible since it only takes a line more for the user to conditional render Irma component.

Hi @imbhargav5 
I totally agree, I  was not clear enough in my previous comment. I am making a change in the API and keeping useGeolocation hook work without any argument. I just wanted to say, that still we can have support for when attribute to allow the user to control the geolocation retrieval. And I understand that this can also be achieved via conditionally rendering a component consuming this hook(your idea). That's where I was wondering. Will the new approach create more complexity?
Check my latest commit and let me know.

imbhargav5:
Ah so when is clubbed with the other geolocation  options? Hmm I think it’s better than the previous example but it’s a very very rare scenario that someone would want to conditionally turn on geolocation. Let’s not have it for now and comment it out. If there is demand for it we can add it back.

simbathesailor:
> Ah so when is clubbed with the other geolocation options? Hmm, I think it’s better than the previous example but it’s a very very rare scenario that someone would want to conditionally turn on geolocation. Let’s not have it for now and comment it out. If there is demand for it we can add it back.
Hi @imbhargav5 
Yes, The scenarios when geolocation is accessed which I faced in my day to day job are:

Add address screens where on click of certain button you fetch geolocation info. Typically these are part of google search auto-suggest inputs. Almost every e-commerce website (where products are based on geolocation info) will have a similar kind of functionality.

Right now, when not passing anything, it behaves like what is desired(less confusion and geolocation enabled by default). I am stretching a little bit here. But my point is that commenting it will not make much difference. By going with this approach we are supporting both the things controlled retrieval of geolocation and uncontrolled retrieval of geolocation. 

I know we have discussed a lot on this, but just letting you know my point of view. I hope you will not mind :). Once again asking you what's your point of view on this?

Thanks

imbhargav5:
I see. No problem then. Let’s add it then. So to be clear it will be together with the geolocation options too right?

simbathesailor:
> I see. No problem then. Let’s add it then. So to be clear it will be together with the geolocation options too right?

Yes, it will be with geolocation option. Earlier I was sending two sets of options one for hook specific and one for geolocation. But now I have made it one.

Issue #103, opened 2019-09-05
New useGeolocation is stuck in an infinite loop.


imbhargav5:
https://dev.react-hooks.org/?path=/story/usegeolocation--basic-example-geolocation-conditional

simbathesailor:
Ohh, I will have a look in sometime.

imbhargav5:
Let's put in a test case to make sure this is covered.

imbhargav5:
Fixed with e5963ccca2aea6ce90b725ad2b916692cd47e60d

PR #111, opened 2019-09-05
Fix infinite loop issue and add test case for it

simbathesailor:
This pr fixes the issue mentioned in issue [https://github.com/imbhargav5/rooks/issues/103](url)

1. Fix the infinite loop issue:
Happening due to the options object being sent as a new reference always and effect running again and again. Now I have spread the options

2. Add test case for the infinite loop issue.


imbhargav5:
I think there should be a better solution to this.

simbathesailor:
Hi @imbharghav
 Any suggestions you have ?How usually people make sure that objects keys are compared and not the object itself. Can think of write a util function or hook to do this task . What are your thoughts ?

imbhargav5:
@simbathesailor Actually this is good. I am going to merge this now.

Issue #93, opened 2019-08-23
useLocalStorage and useSessionStorage perf upgrades


imbhargav5:
- [x] usage of useCallback
- [x] storage event listener is added on each render. Must only be added on mount.

simbathesailor:
Raise a  pull request for this issue [https://github.com/imbhargav5/rooks/pull/128](issue-128)

PR #128, opened 2019-09-09
adding the perf improvements for useLocalstorage and useSessionstorage

simbathesailor:
Add performance improvement mentioned in https://github.com/imbhargav5/rooks/issues/93
 1. usage of useCallback. Done
 2. storage event listener is added on each render. Must only be added on mount. Done.

Let me know if I missed anything ?

### DIFF
--- a/packages/localstorage/src/useLocalstorage.ts
+++ b/packages/localstorage/src/useLocalstorage.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 
 interface StorageHandlerAsObject {
   value: any;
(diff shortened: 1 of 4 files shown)


imbhargav5:
Looks good at first glance. Will review shortly.

imbhargav5:
@simbathesailor Since we are on this. Can we add basic tests for both these hooks using mock tools?

simbathesailor:
> @simbathesailor Since we are on this. Can we add basic tests for both these hooks using mock tools?

Sure, Let me try that !!

imbhargav5:
@simbathesailor So the tests are working without mocks?

simbathesailor:
> @simbathesailor So the tests are working without mocks?

Yes @imbhargav5 !!

PR #133, opened 2019-09-17
useIntersectionObserver

simbathesailor:
Pending Item:

Codesandbox Intersectionobserver example: https://codesandbox.io/s/useintersectionobserver-nvj3f
1. Test cases
2. More practical example in storybook need to be added
3. discuss the API.


simbathesailor:
Hi @imbhargav5 ,
I think this pr can be reviewed , meanwhile I will try to add few test cases. Scrolling and bring an item is view port is quite non-trivial in testing. But as I said I will have a look what all extra test cases I can add.

PR #140, opened 2019-10-02
Add check for SSR

maciekgrzybek:


packages/online/src/useOnline.ts
```diff
@@ -29,12 +29,16 @@ function useOnline(): boolean {
   // we only needs this to be set on mount
   // hence []
   useEffect(() => {
-    window.addEventListener("online", setOnline);
-    window.addEventListener("offline", setOffline);
-    return () => {
-      window.removeEventListener("online", setOnline);
-      window.removeEventListener("offline", setOffline);
-    };
+    if (typeof window === 'undefined') {
```
imbhargav5: We need to put this inside the `getIsOnline` function and return `null` on the server side.

maciekgrzybek:
Hey @imbhargav5 , I made the change, not sure if that's what you've guys meant though. Let me know please.

packages/online/src/useOnline.ts
```diff
@@ -16,7 +19,7 @@ function getIsOnline(): boolean {
  * @returns {boolean} The value of navigator.onLine
  */
 function useOnline(): boolean {
-  const [online, changeOnline] = useState<boolean>(false);
+  const [online, changeOnline] = useState<boolean>(getIsOnline());
```
thibaultboursier: Here, at each render, getIsOnline function will be called. You should pass a callback function to useState. This callback will only be called once.

```js
const [online, changeOnline] = useState<boolean>(() => getIsOnline());
```

imbhargav5: useState with callback? That would mean that the `online` initial value will be a function. Right?

imbhargav5: We can actually do something like this,

```javascript
const onlineInitial = useMemo(() => getIsOnline(), [])
const [online, changeOnline] = useState(onlineInitial);
```

imbhargav5:
@maciekgrzybek Hey sorry it took a bit longer for me to pull this change into the repo, i was working on 3.4.0 which has some really good improvements. Now that is out of the way, we will tackle the stale PRs quickly.

maciekgrzybek:
No worries mate :) just let me know what approach is agreed on :)

imbhargav5:
@maciekgrzybek Yep waiting to hear from @thibaultboursier

packages/online/src/useOnline.ts
thibaultboursier: No. We have the possibility to pass a callback function to useState ([lazy initial state](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state)). React will call this callback function only once.

> "The initialState argument is the state used during the initial render. In subsequent renders, it is disregarded."

The result returned by this callback function will be the new state. 

I made an example: [https://stackblitz.com/edit/react-xs2bu2](https://stackblitz.com/edit/react-xs2bu2)

It's the best way to set an initial state value from a function. It's not the same that
```js
const [online, changeOnline] = useState<boolean>(getIsOnline());
```
In this case, at each render, getIsOnline function will be called. What we want is to have only an initial value from a function call on the first render.

thibaultboursier:
> @maciekgrzybek Yep waiting to hear from @thibaultboursier

@imbhargav5 Answer done :-)

packages/online/src/useOnline.ts
imbhargav5: Nice. Let's do this ? @maciekgrzybek

maciekgrzybek: @thibaultboursier nice one didn't know about it :)

imbhargav5:
@maciekgrzybek The build is failing because `null` isn't assignable to `boolean`. So we will have to change the type of the return value. from `boolean` to `boolean | null` and that should do it.

maciekgrzybek:
Hey @imbhargav5 that makes sense, just fixed it. Any idea why TS was not telling me about the error?

thibaultboursier:
Good job!

PR #142, opened 2019-10-04
feat(use-mouse): improve use mouse

braxtonchristensen:
Added more mouse events than 'mousemove' add support for 'mouseup', 'mousedown' and pass target through.

If this is acceptable I will add unit tests for this hook.

### DIFF
--- a/packages/mouse/src/useMouse.ts
+++ b/packages/mouse/src/useMouse.ts
@@ -1,63 +1,48 @@
 import { useEffect, useState } from "react";
 
 interface MouseData {
-  x: number | null;
-  y: number | null;
-  screenX: number | null;
-  screenY: number | null;
-  pageX: number | null;
-  pageY: number | null;
-  clientX: number | null;
-  clientY: number | null;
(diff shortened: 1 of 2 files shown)


imbhargav5:
Hi @braxtonchristensen Thanks for the PR! We made the useMouse hook to actually just track the mouse position and mouseup and mousedown won't really change the mouse position. So this is probably not relevant to this particular hook. Perhaps we can think of a different hook for this usecase? Also react's next major version (react flare) will have a lot of these events much more easier to manage with event responders, so we probably don't even need to make a hook for this behavior. 
Let me know what you think.

braxtonchristensen:
That sounds awesome, I would love to talk more hook ideas I joined the spectrum channel and it was kind of dead. That GitHub issue the best place? Also I noticed a lot of these hooks could use some tests, what is the best way I can contribute to such an awesome project?

imbhargav5:
Yes you can always create an issue here. I don't check spectrum too often 😓

PR #143, opened 2019-10-06
@rooks/use-modal

capJavert:
`useModal` hook exposes modal logic through react context, also provides `useToggle` hook for controlling modal from other components.

- [x] Implementation
- [x] Tests
- [x] Storybook demo
- [x] README


capJavert:
For some reason useDebounce hooks tests failed. I noticed that they sometimes fail and sometimes don't. Probably some racing condition. For example here it passed on my branch build: https://travis-ci.com/capJavert/rooks/builds/130594989

capJavert:
@simbathesailor converted to TypeScript.

imbhargav5:
@capJavert Looks really good. Generally I feel that rooks is a package meant to be separate from UI as much as possible but I after looking at the code, I am convinced that we can do more with rooks and include more such packages into it. I like the `useModal` approach here since it doesn't include any styling. Does this support `nested` modals by any chance?

capJavert:
@imbhargav5 sure, `nested` modals should be possible. Glad you like it, any chance you can give me ETA for merging this?

imbhargav5:
Will review the code and let you know tonight.

packages/modal/README.md
```diff
@@ -0,0 +1,85 @@
...
+### Installation
+
+```
+npm install --save @rooks/use-modal
+```
+
+### Importing the hook
+
+```js
+import useModal, { ModalProvider, useToggle } from "@rooks/use-modal"
```
imbhargav5: We already have a hook called `useToggle`. Can we name this something else?

packages/modal/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "main": "lib/index.js",
```
imbhargav5: This might be an outdated package.json generation. Can you please take a look at another hook in the `packages` and look at the package.json fields. It was changed fairly recently.

imbhargav5:
> @imbhargav5 sure, `nested` modals should be possible. Glad you like it, any chance you can give me ETA for merging this?

I am going to try and get this merged before Wednesday this week.

packages/modal/README.md
capJavert: Refactored so `useToggle` is not needed. `useModal` registration kinda acts as singleton so only the first call to `useModal` does the registration and other calls just use already created modal context. I think this kind of API is also more straightforward. 

Also `ModalProvider` is now exported as `useModal.Provider`.

packages/modal/package.json
capJavert: Rebased to your `dev` branch. Should be ok now.

packages/modal/README.md
capJavert: Also added this package `packages/rooks`.

imbhargav5:
@capJavert I have been in a dilemma on how best to put this package alongside `rooks` main library. What if we had a package called `rooks-ui` or something where we put in all logic that is more that just behaviour? So `useModalState`, `usePopupState` etc? I know this is taking a touch longer but I don't want to alienate the current users who see `rooks` as a package of utils decoupled from ui. 
So my proposed solution is to create a package for all  ui packages within `/packages/ui` or something. And packages like `packages/modal` won't be added to `rooks` but to `rooks-ui` and we can build a library of hooks like modals, dropdowns, popups etc on it.

imbhargav5:
@simbathesailor Made quite a few pull requests which I had to deny into rooks package because they were UI centric. But with rooks-ui we can put them all in the library I feel.

capJavert:
Sure we could do that. I don't quite understand what you mean by UI centric because you already have hooks like `input`, `select`, `toggle` which if you ask me are also all UI centric in that regard. All in all I don't mind further split but potentially you have even more packages for `@rooks-ui`.

@imbhargav5 so I think next steps would be that I move `modal` to `packages/ui` and adjust package name to `@rooks-ui`.

Then after that we can setup the whole package and add others that were mentioned.

imbhargav5:
Yeah. So `input` only really returns a value and a function. So do selects and toggles. With `modal` we are now entering the `component` territory (Context Provider components or any HOC components for that matter). The coupling will only increase after this point. 
So for now, I am going to release `@rooks/use-modal` separately for now after I review it again and once we have a couple more packages I am thinking of putting them `rooks-ui`. 

PS: I am currently in the process of publishing a book. So I have been a lot slower this month. I am going to review `use-modal` once again and publish this. Thanks for your patience @capJavert

Issue #144, opened 2019-10-07
create useFullscreen?


braxtonchristensen:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. 
I'm always frustrated when I am trying to make an app fullscreen, making it compatible for all browsers, remembering each API.

**Describe the solution you'd like**
a `useFullscreen` hook that will return a fullscreen toggle event that could be attached to an onClick handler to toggle fullscreen.

**Describe alternatives you've considered**
a non react specific library could be used. 

**Additional context**

simbathesailor:
@braxtonchristensen, Thanks for raising this feature request. I found this article, which might be useful.  https://hacks.mozilla.org/2012/01/using-the-fullscreen-api-in-web-browsers/

imbhargav5:
Yes please! Feel free to send in a PR!

PR #150, opened 2019-10-11
Feat/144/add use fullscreen

braxtonchristensen:
#144 
implement `useFullscreen`

Oct-10-2019 22-37-19


packages/fullscreen/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "main": "lib/index.js",
```
imbhargav5: Some of the fields in the package.json are missing. Can you look at the other hooks as example?

packages/fullscreen/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      setIsFullscreen(false);
+      setElement(null);
+
+      Promise.resolve(document[fn.exitFullscreen]()).catch(reject);
+    });
+
+  const toggle = (element?: HTMLElement) =>
+    Boolean(document[fn.fullscreenElement]) ? exit() : request(element);
+
+  return {
```
imbhargav5: Is there anyway we can make the return value an array with as few elements as possible?

braxtonchristensen: Sure, each return value has a use, I assume `toggle` would make it in the most valuable but what about others?

braxtonchristensen: I started the refactor and I feel like the API is not as clean for example, I want one component to toggle fullscreen
`const { toggle } = useFullscreen();`

then I want another to react to it
`const {isFullscreen, onchange} = useFullscreen();`

with array API it might look something like this.

`const [toggle] = useFullscreen()`
then
`const [_toggle, isFullscreen, _isEnabled, onchange] = useFullscreen();`

or worse I need the fullscreen element
`const [_toggle, _isFullscreen, _isEnabled, _onchange, _onerror, _request, _exit, element] = useFullscreen();`

braxtonchristensen: let me know your thoughts.

imbhargav5: cc @simbathesailor @maciekgrzybek

packages/fullscreen/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  for (; i < l; i++) {
+    val = fnMap[i];
+    if (val && val[1] in document) {
+      for (i = 0; i < val.length; i++) {
+        ret[fnMap[0][i]] = val[i];
+      }
+      return ret;
+    }
+  }
+};
```
maciekgrzybek: Can we use forEach here instead of for loop? It would be could to use modern standards.

packages/fullscreen/src/index.ts
maciekgrzybek: I think that all depends on what we want to expose to the API? Do we really need to use literally everything? If so, I don't think that returning that object would be a problem.

packages/fullscreen/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  };
+  const off = (event: string, callback: EventCallback) => {
+    const eventName = eventNameMap[event];
+    if (eventName) {
+      document.removeEventListener(eventName, callback, false);
+    }
+  };
+  const exit = () =>
+    new Promise(function(resolve, reject) {
```
maciekgrzybek: Can you make sure you're consistent with functions? It's better to use arrow functions in callbacks, as the can be anonymous.

packages/fullscreen/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+        resolve();
+      };
+
+      on("change", onFullScreenEntered);
+
+      element = element || document.documentElement;
+      setIsFullscreen(true);
+      setElement(element);
+
+      Promise.resolve(element[fn.requestFullscreen]()).catch(reject);
```
maciekgrzybek: Are these methods going to work for all browsers? Seems like they're all defined in the first part of the code but not being used anywhere? Or am I missing something?

packages/fullscreen/src/index.ts
maciekgrzybek: For example, why do we need to expose the element? Shouldn't that be passed as an argument?

packages/fullscreen/src/index.ts
braxtonchristensen: Good call, thank you for pointing this out.

packages/fullscreen/src/index.ts
braxtonchristensen: In the `for` loop it builds the proper function map for the current browser. For example, if it was running in IE

the map might look something like this
```javascript
{
      requestFullscreen: "msRequestFullscreen",
      exitFullscreen : "msExitFullscreen",
      fullscreenElement: "msFullscreenElement",
      fullscreenEnabled : "msFullscreenEnabled",
      fullscreenchange: "MSFullscreenChange",
      fullscreenerror: "MSFullscreenError"
}
```

this way I am standardizing the API, and I can use the same function names that will map to the proper function without a bunch of conditional checks throughout the code.

braxtonchristensen:
> For example, why do we need to expose the element? Shouldn't that be passed as an argument?

It is an optional argument if you wanted to make the entire app fullscreen you wouldn't pass `request` or `toggle` any arguments this will by default target the root of the application.

If you have multiple things that can go full screen it might be helpful to be able to use this property to know what is fullscreen at the moment.

packages/fullscreen/src/index.ts
braxtonchristensen: I will change it, `for` loops usually have better performance to my knowledge but I agree that it would probably be more readable and since the loop is so small I am sure subscribers wont really feel it.

https://jsperf.com/for-vs-foreach/75

packages/fullscreen/src/index.ts
braxtonchristensen: It is an optional argument if you wanted to make the entire app fullscreen you wouldn't pass request or toggle any arguments this will by default target the root of the application.

If you have multiple things that can go full screen it might be helpful to be able to use this property to know what is fullscreen at the moment.

maciekgrzybek: Hmm, but shouldn't it be a different 'instance' of hook called for each element?

maciekgrzybek: So there will be always only one element (or root) per hook.

packages/fullscreen/src/index.ts
maciekgrzybek: Right didn't notice that, makes perfect sense.

packages/fullscreen/src/index.ts
braxtonchristensen: Good point, I updated it so it will derive the default hook state from document. This will allow the hook to share state across the application in a sense

Component 1: Set Document to full screen
Component 2: Mounts with default state derived from `document`

thoughts on this?

imbhargav5:
We should still go for array destructing to make it easy for us to add more return values later without breaking previous releases. Let's figure out a way to put the most relevant ones at the first few indices in the array.

imbhargav5:
@braxtonchristensen Really sorry about the delay here. Life intervened. Had very little time to maintain rooks until now. Have finally found sometime now. Going to merge this after a few fixes in the next couple of days.

PR #152, opened 2019-10-12
test(mouse): add tests for use mouse

braxtonchristensen:
* test the mouse position is updated when the mouse is moved.

if these tests make sense I could look into adding tests for others. 😄


braxtonchristensen:
Let me know if that is any better.

PR #156, opened 2019-10-15
Update homepage value on each package.json

elrumordelaluz:
Probably should be updated also the `create.js` file or something in the `/template` folder. But before make those changes I would like to know the logic behind the _stories_ like: do you create those manually? always create a `basic example`? etc…


imbhargav5:
@elrumordelaluz It's fine. We can update that in a different PR. Yeah we create the basic stories manually. Is there a better approach?

PR #167, opened 2019-11-25
Fix types for use-outside-click-ref hook

mxstbr:


imbhargav5:
@mxstbr Thank you!! What do you think about this project? :D I have been away from a lot of OSS work because I was focusing on this one the past year.

mxstbr:
It's awesome, thank you for making it! :100:

PR #169, opened 2019-11-29
Update use-outside-click-ref types to add missing argument

mxstbr:


imbhargav5:
Let's rename the variable to `when`? All the other hooks have the name `when` for behaviour. So it's more consistent. Actually now that I think of it `isEnabled` is perhaps the best name.

mxstbr:
Updated!

PR #170, opened 2019-11-30
@rooks/use-time

Stupidism:
# Description
- Add a hook to track time tick

## Snapshot 

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist for Author
- [x] Tests written where needed
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PS: I will add another hook to do `countdown` based on this hook.

image


package.json
```diff
@@ -36,6 +36,7 @@
     "@babel/preset-env": "7.5.5",
     "@babel/preset-react": "7.0.0",
     "@testing-library/react": "8.0.1",
+    "@testing-library/react-hooks": "^3.2.1",
```
Stupidism: I used another testing-library for hooks so we don't need to write mock components to represent the results of hooks any more.

package.json
```diff
@@ -54,6 +55,9 @@
     "mini-css-extract-plugin": "0.8.0",
     "ora": "3.0.0",
     "raf": "3.4.1",
+    "react": "16.9.0",
```
Stupidism: `@testing-library/react-hooks` used a feature in `react@^16.9`.
It doesn't affect the hooks themselves but only needs `react` to be over `16.9` in the dev environment.
I'm not very sure if this is a good way to do it.

Stupidism:
Somehow, `useWindowScroll` failed in my pull request.
image

Stupidism:
It looks like the `GITHUB_TOKEN` is missing.
According to the [docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#permissions-for-the-github_token), the `GITHUB_TOKEN` includes the permission of adding comments on PRs.

image

imbhargav5:
Hey @Stupidism, Thanks for submitting this PR!! 
Is this hook similar to `useInterval`? It's not entirely clear to me what this hook is aiming to do.

Stupidism:
@imbhargav5 You're right. I find that I can write a useCountdown hook based on useInterval instead of this useTime.

Issue #172, opened 2019-11-30
Greetings on first PR github action is failing.


imbhargav5:
**Describe the bug**
The github action which sends a message on first PR submission from a user is failing.

PR #204, opened 2020-01-27
Add param to listen to mousedown instead of click

### LINKED ISSUE #199: mousedown support for use-outside-click(-ref)
We are using D'n'D list with ability to modify elements right in the list using double click. Click on Cancel or outside of item. However when item is in edit mode can press left mouse button on any other item in list, drag it and item in edit mode won't be converted to view mode.

Currently these hooks [use `click` event](https://github.com/imbhargav5/rooks/blob/1e950073f9a14023c36923736c70432761c3cc0a/packages/outside-click/src/useOutsideClick.ts#L31). In our case it will be preferable to use `mousedown` to fix this problem.
Options are:
1) Replace `click` with `mousedown` for everyone (not sure if this won't break existing usages)
2) Add third parameter to hook that will switch from `click` to `mousedown` if `true`
3) Same as above but replace `when` parameter with object `{isEnabled: boolean, mousedown: boolean}`, similar to `addEventListener`'s third parameter.

We'll patch events locally or write own hook implementation.

**Additional context**
While switch from `click` to `mousedown` fixes described problem with mouse events, it looks like similar problem can be reproduced with touch (tested Windows 10, Edge 79 (Chromium)). Change of event will bring asymmetry between touch and mouse in this aspect.

Switch to `pointerdown` in future will cover all cases. [Pointer Events API support](https://caniuse.com/#feat=pointer)

I can submit PR if any of 3 options works for you.

IllusionMH:
Fixes #199

After few attempts (to overengineer things) it looks like it shouldn't be a big deal to just reassign events when `listenMouseDown` changes, and now it's inline with `when` behavior.

I haven't found tests/story/readme description for `when` parameter, do I need to add these for `listenMouseDown` everything in scope of this PR?

Looks like failure related to #172 and not PR changes.

UPD. Looks like I've missed that this repo already have `@testing-library/react`. Tried to play with tests https://codesandbox.io/s/bold-flower-k3b5u so shouldn't be too hard to add them to this repo.

### DIFF
--- a/packages/outside-click-ref/README.md
+++ b/packages/outside-click-ref/README.md
@@ -35,3 +35,18 @@ function Demo() {
 
 render(<Demo/>)
 ```
+
+### Arguments
+
+| Arguments       | Type      | Description                                                                                                                                      | Default value |
+| --------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------ | ------------- |
+| ref             | React ref | Ref whose outside click needs to be listened to                                                                                                  | N/A           |
+| handler         | function  | Callback to fire on outside click                                                                                                                | N/A           |
+| when            | boolean   | A boolean which activates the hook only when it is true. Useful for conditionally enable the outside click                                       | true          |
+| listenMouseDown | boolean   | A boolean which switches the hook to listen to `mousedown` event instead of `click`. Useful when user might start drag instead of regular clicks | false         |
+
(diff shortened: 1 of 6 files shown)


IllusionMH:
Added tests and updated readme with list of arguments (manually). I think it's ready for review.

Please let me know if anything is missing in this PR.

PR #257, opened 2020-10-01
Change effect to use use-effect

imbhargav5:
cc @weotch Let me know!


weotch:
Yeah, that looks right to me!

Issue #313, opened 2020-11-27
Show contributors in home page


imbhargav5:
- We basically need to populate a file called contributors.md or something like that at /packages/docusaurus/docs/contributors.md and use it in our homepage. 
- Update the file in our PR release workflow

Avi98:
Hi @imbhargav5 can I work on this

imbhargav5:
For sure!

Avi98:
Hi @imbhargav5  do you want me to maintain a hardcoded file called contributor.md or to write a script using that extracts the contributors from .all-contributorsrc.

imbhargav5:
So we need to generate a contributors.md inside the docusaurus package using all contributors cli within our script. 

We can use the same script to update our readme.

PR #324, opened 2020-12-13
contributors file added in docs

Avi98:
Related to #313

- [x] added contributors.md inside docs
- [ ] update GitHub workflow update-title-card

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -4,7 +4,8 @@
   "repoType": "github",
   "repoHost": "https://github.com",
   "files": [
-    "README.md"
+    "README.md",
+    "./packages/docusaurus/docs/contributors.md"
   ],
   "imageSize": 100,
   "commit": false,
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "readme-to-docs": "lerna exec --scope=@rooks/* --scope=rooks --concurrency 1 --stream -- 'node ../../helpers/docs/index.js'",
     "bs": "lerna bootstrap",
     "prebuild": "lerna clean -y && yarn",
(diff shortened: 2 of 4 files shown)


imbhargav5:
So, the important part of this issue is to make sure we run this everyday as a GitHub action. So you can take a look at a couple of other GitHub workflows we already have that run once a day. 

The idea is that we run this task once every 1 or 2 days and make sure this contributors list is always upto date. What say?

Avi98:
> So, the important part of this issue is to make sure we run this everyday as a GitHub action. So you can take a look at a couple of other GitHub workflows we already have that run once a day.
> 
> The idea is that we run this task once every 1 or 2 days and make sure this contributors list is always upto date. What say?
> 
> So similar to update-title-card, we will have to create another GitHub workflow that runs once every 2 days.



> So, the important part of this issue is to make sure we run this everyday as a GitHub action. So you can take a look at a couple of other GitHub workflows we already have that run once a day.
> 
> The idea is that we run this task once every 1 or 2 days and make sure this contributors list is always upto date. What say?
> 
> So similar to update-title-card, we will have to create another GitHub workflow that runs once every 2 days.

Yes @imbhargav5 , will try to complete by tomorrow EOD.
Thanks a lot for looking into my PR

imbhargav5:
I must be thanking you. Thanks for this work man!

Avi98:
Hi @imbhargav5, I'm new to github actions. Can I add a job to existing workflow ```update-title-card```  as it already runs after every two days. The job will be something similar to this

```yml
- name: update-contributors-list
        run: yarn run contributors:generate
```

imbhargav5:
@Avi98 For sure. Sorry I responded late.

PR #334, opened 2020-12-30
update use did update

zeritte:
In the past, useDidUpdate didn't run the cleanup function of the callback.

```
useDidUpdate(() => {
   API.subscribe(userID)
   ()  => API.unsubscribe(userID)
}, [userID])
```

Before the PR, unsubscribe function is never called. With the PR, it will be called properly


zeritte:
https://github.com/streamich/react-use/blob/master/src/useUpdateEffect.ts similar approach

imbhargav5:
Actually that was intentional. I wanted it to look very similar to componentDidUpdate which doesn't have a cleanup. Perhaps we can do another hook like `useUpdateEffect`or something? Would love if you want to do a PR for that instead!

zeritte:
I created useUpdateEffect but I think we should have merged it with useDidUpdate instead. If a user wants to use a cleanup, he can. If he doesn't want to use a cleanup like componentDidUpdate, he doesn't have to.

For useUpdateEffect function, svg file and codesandbox.io example are missing

imbhargav5:
Yeah the only reason why I am thinking of a different hook is because I don't want to cause breaking changes if someone is using useDidUpdate with async. That would break if we put in a cleanup since async functions return a promise but cleanup needs to be a function.

zeritte:
OK I see your point

zeritte:
@imbhargav5 hey, did you have a chance to look at this?

imbhargav5:
Hi I am resting right now. Will be back in a week. Thanks for your
patience.

imbhargav5:
Going to add test cases and merge today!

imbhargav5:
@all-contributors please add @zeritte for code.

PR #342, opened 2021-01-21
Ensure visibility check doesn't throw warnings on SSR

weschristiansen:
Use the existing isomorphic hook to prevent warnings


imbhargav5:
@weschristiansen Thanks for this PR! Looks good to me!

PR #344, opened 2021-01-24
Fix/ssr use fullscreen

### LINKED ISSUE #260: [useFullscreen] useFullscreen is not SSR-friendly
**Describe the bug**
`useFullscreen` directly accesses `document` so can't be used with server-side rendering. Errors produced are e.g.

```
ReferenceError: document is not defined
```

**To Reproduce**
Use SSR with a component that calls `useFullscreen`:

```js
const { renderToString } = require('react-dom/server');
const useFullscreen = require('@rooks/use-fullscreen');

renderToString(() => {
  useFullscreen();
  return <></>;
});
```

**Expected behavior**
`useFullscreen` should probably do a 2-pass render, with the first pass (server-side render and browser initial render) having fullscreen support off and a second (browser only) if fullscreen support is detected for the browser.

maciekgrzybek:
PR to fix #260


imbhargav5:
Looks good at first glance. Will take a look! Thanks a ton @maciekgrzybek

packages/shared/useFullscreen.ts
```diff
@@ -83,7 +83,11 @@ type FullscreenApi = {
  * useFullscreen
  * A hook that helps make the document fullscreen
  */
-function useFullscreen():FullscreenApi{
+function useFullscreen(): FullscreenApi | undefined {
+  if (typeof window === 'undefined') {
+    return;
```
gpoole, line 88: I think it'd be nice to return a negative match here rather than `undefined` to avoid users having to handle the SSR case specially e.g.

```
{
  isEnabled: false,
  isFullscreen: false,
  // ... etc
}
```

gpoole (commented):
Thanks @maciekgrzybek!

I'm not totally sure but from skimming through I get the feeling that this isn't going to trigger a `setState` after the first render in all cases so it won't be guaranteed to do the second pass in the browser after rehydration. I think there might need to be some additional logic to do that. Correct me if I'm wrong though?

packages/shared/useFullscreen.ts
maciekgrzybek: Hmm I get your point, but not sure how would that be beneficial for the user? Will still need to do a check what was returned from the hook. With returning `undefined` it's clearly stated - "Hey, it's SSR, it won't work on the server". I'm happy to discuss more though :) What's your opinion on this @imbhargav5 ?

P.S. It definitely works with Next.js, and from my experience with SSR it will trigger `setState` when coming to the browser.

imbhargav5: Running tests now @maciekgrzybek . Going to nail this in the next 2 hours. This hook is by far the most insecure hook we ever made :D

imbhargav5: Yeah. I think it's best we return some falsy values on the server. Will do that now.

imbhargav5: cc @gpoole

gpoole: > not sure how would that be beneficial for the user?
The benefit I see would be that during SSR the user's code can follow the same path as when fullscreen support is not present, rather than needing to write a special case just for SSR.

> from my experience with SSR it will trigger setState when coming to the browser
The issue I'm thinking of would be that to work correctly the 2-pass steps in the client should be: render once with fullscreen support off (matching the server state), then render a second time if fullscreen support is detected (2nd pass). If you skip the first step immediately render in the fullscreen supported state, which is what I think is happening here, you may have a mismatch with the SSR output during the first hydration pass and could run into issues.

imbhargav5:
@all-contributors please add @AdrienLemaire for code.

packages/shared/useFullscreen.ts
imbhargav5: Yeah. We can fail silently on the server as a convenience with falsy values but doing it in 2 steps is needlessly complicated for the client side only scenario.

imbhargav5: Also, as I am skimming through the code, I find the `onChange` and `onError` functions pretty difficult to work with and to refactor.

gpoole: Yeah I understand. The problem will be that not doing both passes will cause DOM mismatch problems when using SSR if different output is returned depending on whether fullscreen is available. It definitely is an unnecessary step for the client only scenario unfortunately, yes, but I'm not aware of a better way that avoids the issue.

gpoole: One thing I can think of is that it is an additional performance hit but also one that I would have thought was relatively minor for the majority of users. Potentially a flag to disable 2-pass for users concerned about performance could cover that?

imbhargav5: It's shaping out well now. Should be able to work solidly from now on.

imbhargav5: @gpoole Because useFullscreen has no purpose on the server, we shouldn't do anything more than silently failing on the server and optimise best for the client side where it is meant to be used. 
I feel like a lot of devs can simply use `next/dynamic` or `react-no-ssr` or something quite easily and wrap a component around and make it work client only for fullscreen. 
So it's best we make this hook work flawlessly on client and make it not too bad on server.

imbhargav5:
@maciekgrzybek @gpoole Works amazingly well now :D 

[Image from Gyazo](https://gyazo.com/cbb40f1388bbcbfa756dba9db9ee2f3a)

maciekgrzybek:
Awesome :) (BTW, that image doesn't work for me :) ) 
Should that be merged now?

imbhargav5:
@maciekgrzybek @gpoole @simbathesailor  Works amazingly well now :D 

https://i.gyazo.com/cbb40f1388bbcbfa756dba9db9ee2f3a.gif


Check out the demo here https://q4hpf.csb.app/

imbhargav5:
I am going to merge this and push out a canary.

imbhargav5:
v4.8 is out. And fullscreen hook works well now. cc @maciekgrzybek @gpoole @danielholmes 

Here is a sandbox you can try https://codesandbox.io/s/immutable-wave-m87bx?file=/src/App.js

maciekgrzybek:
Nice work :)

imbhargav5:
:D It was super painful to work on this hook because everything was mangled. But I figured out most of it in the end. I silently deprecated `onChange` and `onError` in the output form but put an alternative in the input form. It won't throw errors and still does a fair job although buggy.

However, `onChange` in argument form works superbly well.

gpoole:
Great, thanks for that @maciekgrzybek @imbhargav5! Love the rooks packages, such a great reference and so many useful tools.

packages/shared/useFullscreen.ts
gpoole: 👍 understood, that's fair enough

PR #346, opened 2021-01-25
Avoid unnecessary effect calls due to use-fullscreen

AdrienLemaire:
useFullscreen returns new functions whenever it is being called. That means that every useEffect that uses a function from useFullscreen will be re-executed.
This created issues for us (duplicate notifications sent from said useEffect).

The solution is to return a memoized function by using useCallback.
Haven't looked at other hooks, but I'm afraid this problem is not only affecting this hook...


imbhargav5:
Going to get to this after I get home tonight. Thanks!

maciekgrzybek (approved):

imbhargav5:
@maciekgrzybek Is this tested?

maciekgrzybek:
No I haven't tested it

imbhargav5:
Thanks @AdrienLemaire !!

PR #384, opened 2021-02-13
Add prettier pre-commit hook

### LINKED ISSUE #380: Setup prettier
We have a `prettier` installed in the project. It would make sense to run it on pre-push hook.

I feel like we should use husky to handle the hook and run prettier before the push.

Run prettier in the pipelines. Although I think it would be better to set up formatting before pushing it.

yakkomajuri:
When the user runs `git commit` this will trigger prettier in "write" mode so it will automatically lint the files according to the prettier configuration the repo already has at `.prettierrc`.

Also moved prettier to `devDependencies` but feel free to disagree with me there.

Close #380

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -2,18 +2,19 @@ name: Version bump
 
 on:
   push:
-    tags: ["v*"]
+    tags: ['v*']
 
 jobs:
   build:
-    runs-on: ubuntu-latest  
+    runs-on: ubuntu-latest
     env:
-      ALGOLIA_API_KEY:   ${{ secrets.ALGOLIA_API_KEY }}
+      ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+      IS_GH_ACTION: 1
     steps:
-      - name: "Check if user has write access"
-        uses: "lannonbr/repo-permission-check-action@2.0.0"
+      - name: 'Check if user has write access'
+        uses: 'lannonbr/repo-permission-check-action@2.0.0'
         with:
-          permission: "write"
+          permission: 'write'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Checkout the tag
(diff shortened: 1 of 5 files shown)


maciekgrzybek (approved):
Looks good to me. @imbhargav5 ?

imbhargav5:
Thanks a ton! @yakkomajuri 

Yeah `devDependencies` normally make sense but in yarn workspaces they aren't very different from `root` `dependencies`. So let's have them in `dependencies` if possible. 

Also, we have some automated commits that occur in GitHub actions, can we disable `lint-staged` to fire on such commits or is it already taken care of?

yakkomajuri:
@imbhargav5 Shouldn't really pose problems to the bots, but nevertheless by default it wouldn't lint .lock files and I've now added a prettier rule to ignore package.json too, given that and `yarn.lock` are usually what dependabot targets.

Moved packages to dependencies and sorted them too.

yakkomajuri:
Let me know if you want `11a0ddb` (test commit) reverted, I essentially just ran prettier on it.

imbhargav5:
@yakkomajuri I am only thinking of this particular piece of code https://github.com/imbhargav5/rooks/blob/master/.github/workflows/version-bump.yml#L92. I don't want it to ever fail haha. If we can somehow configure lint-stage to not run in GitHub Actions env that would solve my paranoia. 

Also thank you for patiently addressing the fixes so far!

yakkomajuri:
Oh I mean if you really want I can set an env var on that workflow and check for it on the pre commit hook, but `prettier --write` doesn't fail (i.e. it's not like eslint), it just formats your files according to the spec. As of right now, it's configured to do that to files with the following extensions:

`html,js,ts,tsx,json,yml,css,scss`

.github/workflows/version-bump.yml
```diff
@@ -2,18 +2,19 @@ name: Version bump
...
+    tags: ['v*']
 
 jobs:
   build:
-    runs-on: ubuntu-latest  
+    runs-on: ubuntu-latest
     env:
-      ALGOLIA_API_KEY:   ${{ secrets.ALGOLIA_API_KEY }}
+      ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+      IS_GH_ACTION: 1
```
yakkomajuri, line 12: here's the relevant diff

imbhargav5:
Yeah. I wanted to be extra careful. Thanks for this @yakkomajuri !! Going to merge this.

imbhargav5:
@all-contributors please add @yakkomajuri for code.

PR #409, opened 2021-03-30
Add `useMediaMatch` hook to indicate if a media query is currently matched

mrvisser:
Hi folks, love your library!

I had a need to know when a page switched to `print` media  and back, so I created a hook that toggles a boolean based on a `window.matchMatch(<media query>)` listener.

I think we could do something more powerful, such as a `use-media-match-listener` where you specify a listener function, but I thought that this one would be the most common case, anyway.

There's an example usage in the README.md. Here's a couple quick peeks:

```javascript
// `true` if rendered for printing, `false` otherwise
const isPrintMedia = useMediaMatch('print');

// `true` if viewport is smaller than 600px, `false` otherwise
const isNarrowWidth = useMediaMatch('(max-width: 600px)');
```

Let me know if I have everything sorted with the PR or if you want tweaks.

Thanks for reviewing!


packages/shared/useMediaMatch.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  useEffect(() => {
+    const listener = (ev: MediaQueryListEventMap['change']) =>
+      setMatches(ev.matches);
+    matchMedia.addEventListener('change', listener);
+    return () => matchMedia.removeEventListener('change', listener);
+  }, [matchMedia]);
+
+  return matches;
+}
```
maciekgrzybek, line 35: Hmm if I understand the use cases correctly, I don't think you need to setup `matchMedia` as a state:
```suggestion
  const matchMedia = window.matchMedia(query);
  const [matches, setMatches] = useState(() => matchMedia.matches);

  useEffect(() => {
    const mm = window.matchMedia(query);
    setMatches(mm.matches);
  }, [query]);

  useEffect(() => {
    const listener = (ev: MediaQueryListEventMap['change']) =>
      setMatches(ev.matches);
    matchMedia.addEventListener('change', listener);
    return () => matchMedia.removeEventListener('change', listener);
  }, [matchMedia]);

  return matches;
}
```

packages/shared/useMediaMatch.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns Whether or not the media query is currently matched.
+ */
+function useMediaMatch(query: string): boolean {
+  if (typeof window === 'undefined') {
+    console.warn('useMediaMatch cannot function as window is undefined.');
+    return false;
+  }
+
+  const [matchMedia, setMatchMedia] = useState(() => window.matchMedia(query));
+  const [matches, setMatches] = useState(() => matchMedia.matches);
```
maciekgrzybek, line 19: Also it would be good idea to pass the types here :)

packages/shared/useMediaMatch.ts
mrvisser: My understanding is that this hook function gets called potentially a lot. I assume running `window.matchMedia` could have some cost to it. For example it resolves a result set that has `matches: boolean`, so the browser would be actively running some match logic potentially very often. I thought it would be best to only instantiate a `matchMedia` instance when the query string changes rather than in a potentially tight loop.

Do you still think we should make this change?

packages/shared/useMediaMatch.ts
mrvisser: It infers it from `matchMedia.matches` but I can pass some too.

packages/shared/useMediaMatch.ts
mrvisser: Also in your version of the code, you're reacting on dependency `[matchMedia]` in the second useEffect hook which will now change every re-render with your suggested change.

imbhargav5:
Will go through this in a bit. Thanks @mrvisser 🚀

maciekgrzybek:
@mrvisser you are probably right :) I might have misunderstood the problem, I'll try to recreate that on my machine :) one way or another, good PR thanks :)

imbhargav5:
Going well to get this merged during the weekend.

imbhargav5:
Apologies @mrvisser for the delay. Got super sick. Need a bit more time. I understand that waiting is a total pain but please bear with me.

imbhargav5:
The PR looks solid. And I appreciate the tests very much.

packages/shared/useMediaMatch.ts
imbhargav5: @mrvisser Any chance we can do it with just a `useMemo` instead of `useState`? Sorry I joined this discussion a touch late.

mrvisser: Oh yea, I think so and it will be a lot cleaner. Will follow up on that.

mrvisser:
> Apologies @mrvisser for the delay. Got super sick. Need a bit more time. I understand that waiting is a total pain but please bear with me.

> The PR looks solid. And I appreciate the tests very much.

No reason to be sorry, hope you're feeling better! I've followed up on your feedback -- thanks!

imbhargav5:
@mrvisser Sweet. I have just need three more things I need from you before I merge this PR. 

1. Let's make sure that we still immediately track match when `matches` memo value changes . 
 ```jsx
 useEffect(() => {
    setMatches(matchMedia.matches);
    const listener = (ev: MediaQueryListEventMap['change']) =>
      setMatches(ev.matches);
    matchMedia.addEventListener('change', listener);
    return () => matchMedia.removeEventListener('change', listener);
  }, [matchMedia]);
 ```
2. Remove the `await` before `act` in the test spec you wrote. That was giving a warning.
And finally :rocket:
3. Please add this package to `package/rooks/index.ts` file

imbhargav5:
Sweet! @mrvisser . Looks like all the changes I asked for are done.

imbhargav5:
@all-contributors please add @mrvisser for code.

mrvisser:
> 1. Let's make sure that we still immediately track `matches` when `matchMedia` memo value changes .

Good catch. Fixed and insulated it with a test. Thanks!

imbhargav5:
Super! Thanks @mrvisser for the great work!

PR #411, opened 2021-03-30
Add the documentElement fallback when element is not there

### LINKED ISSUE #404: fullscreen doesn't work anymore
**Describe the bug**
calling `request()` doesn't set the screen fullscreen anymore, when it did before.

**Additional context**
While debugging the code, it seems like `document.requestFullscreen` doesn't exist anymore on latest Chrome versions (currently on v89).
Executing `document.body.requestFullscreen()` does work though. Is chrome expecting an Element instead of a document now ?

simbathesailor:
Fix #404 

 Added fallback document.documentElement when element is not present

### DIFF
--- a/packages/shared/useFullscreen.ts
+++ b/packages/shared/useFullscreen.ts
@@ -1,6 +1,6 @@
-import { useState, useCallback, useMemo, useRef } from "react";
-import { useDocumentEventListener } from "./useDocumentEventListener";
-import { warning } from "./warning";
+import { useState, useCallback, useMemo, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


imbhargav5:
Hard to read what's going on in the PR. Is it because of prettier that there are so many changes?

simbathesailor:
Yes 

> Hard to read what's going on in the PR. Is it because of prettier that there are so many changes?

Yes , the only change is one line 
```
const finalElem = element || document.documentElement;
```

This happened due ti prettier and i am not sure how do we fix that.

PR #415, opened 2021-04-06
fix(usefullscreen): change isenabled fullsrceen property

### LINKED ISSUE #410: useFullscreen : isEnabled is not relevant
**Describe the bug**
In the useFullscreen hook, the `isEnabled` property is defined with `Boolean(element)`. But `element` is defined by the document property `fullscreenElement`, which is null if fullscreen is not open.

So `isEnabled` is `false` by default as describe in this documentation : https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API/Guide#other_information

**To Reproduce**
install and use @rooks/use-fullscreen as describe in the documentation : https://react-hooks.org/docs/use-fullscreen/

**Expected behavior**
`isEnabled` should be used as a flag to check if fullscreen is allowed / supported

**Desktop (please complete the following information):**
 - all

**Smartphone (please complete the following information):**
 - all

**Additional context**
This issue should be resolved with changing `isEnabled` declaration from `Boolean(element)` to `Boolean(document[fullscreenControls.fullscreenEnabled])`

thodubois:
Change property `isEnabled` to match mdn documentation.
fix #410

### DIFF
--- a/packages/shared/useFullscreen.ts
+++ b/packages/shared/useFullscreen.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useMemo, useRef } from 'react';
+import { useState, useCallback, useRef } from 'react';
 import { useDocumentEventListener } from './useDocumentEventListener';
 import { warning } from './warning';
 
@@ -206,7 +206,7 @@ function useFullscreen(
   );
 
   return {
-    isEnabled: Boolean(element),
+    isEnabled: Boolean(document[fullscreenControls.fullscreenEnabled]),
     toggle,
     onChange: onChangeDeprecated,
     onError: onErrorDeprecated,


packages/shared/useFullscreen.ts
```diff
@@ -206,7 +206,7 @@ function useFullscreen(
   );
 
   return {
-    isEnabled: Boolean(element),
+    isEnabled: Boolean(document[fullscreenControls.fullscreenEnabled]),
```
imbhargav5, line 209: Is there a merit to doing this over the existing `element` variable? 🤔 

cc @simbathesailor

thodubois: Yes, as it is describe here : https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#properties 

The element only tells you if fullscreen mode is activated (as `isFullscreen` is meant to)
```
Document.fullscreenElement / ShadowRoot.fullscreenElement
    The fullscreenElement property tells you the Element that's currently being displayed in full-screen mode on the DOM (or shadow DOM). If this is null, the document (or shadow DOM) is not in full-screen mode.
```

`fullscreenEnabled` tells you if fullscreen is supported or not
```
document.fullscreenEnabled
    The fullscreenEnabled property tells you whether or not it is possible to engage full-screen mode. This is false if full-screen mode is not available for any reason (such as the "fullscreen" feature not being allowed, or full-screen mode not being supported).
```

simbathesailor: Yes I agree with thodubois , document.fullscreenEnabled should be used instead of Document.fullscreenElement /  ShadowRoot.fullscreenElement. That's what I can understand from docs

thodubois: It seems it was like this before:
https://github.com/imbhargav5/rooks/blob/0180335c454c035c586246df10c7d9ad12f0275b/packages/shared/useFullscreen.ts#L150

imbhargav5: Sounds good!

imbhargav5:
@all-contributors please add @thodubois for code.

PR #475, opened 2021-05-05
fix touch bug, update types, and add tests for useOutsideClick

ramblinjan:
- Adds tests for useOutsideClick (100% file test coverage)
- Adds optional generic parameter for the ref type
- Fixes the handler event type to include touch events
- Adds tests that include both touch and click events. Initially the touch tests were failing so this also:
- Fixes bug with `ontouchstart` being specified where `touchstart` was expected (for some reason typescript doesn't call this out but when the hook is copied into my own application it did -- I wonder if the tsconfig for this app needs updated?)

### DIFF
--- a/src/hooks/useOutsideClick.ts
+++ b/src/hooks/useOutsideClick.ts
@@ -9,9 +9,9 @@ import { useEffect, useRef, useCallback } from 'react';
  * @param handler Callback to fire on outside click
  * @param when A boolean which which activates the hook only when it is true. Useful for conditionally enable the outside click
  */
-function useOutsideClick(
-  ref: MutableRefObject<HTMLElement>,
-  handler: (e: MouseEvent) => any,
+function useOutsideClick<T extends HTMLElement = HTMLElement>(
+  ref: MutableRefObject<T | null>, // initially html element refs will be null
+  handler: (e: MouseEvent | TouchEvent) => any, // click | ontouchstart
   when: boolean = true
 ): void {
   const savedHandler = useRef(handler);
--- a/src/hooks/useOutsideClick.ts
+++ b/src/hooks/useOutsideClick.ts
@@ -29,11 +29,11 @@ function useOutsideClick<T extends HTMLElement = HTMLElement>(
   useEffect(() => {
     if (when) {
       document.addEventListener('click', memoizedCallback);
-      document.addEventListener('ontouchstart', memoizedCallback);
+      document.addEventListener('touchstart', memoizedCallback);
 
       return () => {
         document.removeEventListener('click', memoizedCallback);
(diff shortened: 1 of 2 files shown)


imbhargav5:
Yo @ramblinjan ! Thanks a ton for this. Will get to this today 🤘🤘

src/hooks/useOutsideClick.ts
```diff
@@ -29,11 +29,11 @@ function useOutsideClick(
   useEffect(() => {
     if (when) {
       document.addEventListener('click', memoizedCallback);
-      document.addEventListener('ontouchstart', memoizedCallback);
```
imbhargav5, line 32: Wow. Great catch! @ramblinjan

imbhargav5 (commented):
This is great @ramblinjan. Do you think you can update another hook that is almost exactly like this? Your tests are interesting and I think you would be best suited to update it as well. 

Would appreciate that very much!! 💛

imbhargav5:
Build seems to be failing with Typescript though. That might be something we should look at.

imbhargav5:
Also, please commit the next couple of commits using `yarn cz` instead of `git commit` and that will help us come up with a commit message that semantic-release can parse and automatically publish a patch release. 

So after `yarn cz`, select `fix` and then type in `outside-click` after which you can type 'fixes types and touch events' and the commit message will be generate like so `fix(outside-click): fix types and touch events`.  And this fix commit will inform semantic-release to do a patch release for us.

binajmen:
> Build seems to be failing with Typescript though. That might be something we should look at.

Is it fixed? Still have a Typescript error:

Screenshot 2022-03-05 at 11 26 38

```
No overload matches this call.
  Overload 1 of 2, '(type: keyof DocumentEventMap, listener: (this: Document, ev: MouseEvent | UIEvent | Event | ClipboardEvent | AnimationEvent | ... 13 more ... | WheelEvent) => any, options?: boolean | ... 1 more ... | undefined): void', gave the following error.
    Argument of type '"ontouchstart"' is not assignable to parameter of type 'keyof DocumentEventMap'.
  Overload 2 of 2, '(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions | undefined): void', gave the following error.
    Argument of type '(e: MouseEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
      Type '(e: MouseEvent) => void' is not assignable to type 'EventListener'.
        Types of parameters 'e' and 'evt' are incompatible.
          Type 'Event' is missing the following properties from type 'MouseEvent': altKey, button, buttons, clientX, and 21 more.ts(2769)
```

PR #485, opened 2021-05-13
feat: add usePreventBodyScroll hook

iampika:
Prevent Body Scroll

### DIFF
--- a/helpers/hooks-list.json
+++ b/helpers/hooks-list.json
@@ -158,7 +158,7 @@
     },
     {
       "name": "usePreventBodyScroll",
-      "description": "Prevent Body Scroll"
+      "description": "A react hook to prevent body scroll. Optionally takes a condition when which disables body scroll only when the condition is true."
     },
     {
       "name": "usePrevious",
--- a/helpers/hooks-list.json
+++ b/helpers/hooks-list.json
@@ -157,6 +157,10 @@
       "description": "A hook that can track a click event outside a ref. Returns a callbackRef."
     },
     {
+      "name": "usePreventBodyScroll",
+      "description": "Prevent Body Scroll"
+    },
+    {
       "name": "usePrevious",
       "description": "Access the previous value of a variable with this React hook"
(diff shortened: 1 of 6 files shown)


src/hooks/usePreventBodyScroll.ts
```diff
@@ -0,0 +1,33 @@
+import { useIsomorphicEffect } from './useIsomorphicEffect';
+
+/**
```
imbhargav5, line 3: This JSDoc documentation should be present right above the function `usePreventBodyScroll`. So at line number 18.

src/hooks/usePreventBodyScroll.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type Options = {
+  when: boolean;
+};
+
+const defaultOptions = {
+  when: true,
+};
+
+function usePreventBodyScroll(options: Options = defaultOptions) {
```
imbhargav5, line 19: Let's use `const`

src/hooks/usePreventBodyScroll.ts
```diff
@@ -0,0 +1,33 @@
...
+  const { when } = options;
+
+  useIsomorphicEffect(() => {
+    if (when) document.body.style = 'overflow: hidden';
```
imbhargav5, line 23: Let's change this to `if(when && typeof document !== 'undefined')`

helpers/hooks-list.json
```diff
@@ -156,6 +156,10 @@
       "name": "useOutsideClickRef",
       "description": "A hook that can track a click event outside a ref. Returns a callbackRef."
     },
+    {
+      "name": "usePreventBodyScroll",
+      "description": "Prevent Body Scroll"
```
imbhargav5, line 161: Let's change the description everywhere to 'A react hook to prevent body scroll. Optionally takes a condition `when` which disables body scroll only when the condition is true. '

PR #486, opened 2021-05-15
feat: add useLifecycleLogger hook with tests

iampika:


src/hooks/useLifecycleLogger.ts
```diff
@@ -0,0 +1,42 @@
+import type { EffectCallback } from 'react';
+import { useRef, useEffect } from 'react';
+
+const useEffectOnce = (effect: EffectCallback) => {
```
imbhargav5, line 4: We already have these hooks right? Let's directly use them instead of creating them.

src/hooks/useLifecycleLogger.ts
```diff
@@ -0,0 +1,42 @@
...
+    isFirst.current = false;
+
+    return true;
+  }
+
+  return isFirst.current;
+}
+
+const useUpdateEffect: typeof useEffect = (effect, deps) => {
+  const isFirstMount = useFirstMountState();
```
imbhargav5, line 21: We have a `useDidMount` which we can use here.

src/hooks/useLifecycleLogger.ts
```diff
@@ -0,0 +1,42 @@
...
+  useEffectOnce(() => {
+    console.log(`${componentName} mounted`, ...rest);
+
+    return () => console.log(`${componentName} unmounted`);
+  });
+
+  useUpdateEffect(() => {
+    console.log(`${componentName} updated`, ...rest);
+  });
+};
```
imbhargav5, line 40: Let's also use `useWillUnmount` hook here.

src/hooks/useLifecycleLogger.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    if (!isFirstMount) {
+      return effect();
+    }
+  }, deps);
+};
+
+const useLifecycleLogger = (componentName: string = 'Component', ...rest) => {
+  useEffectOnce(() => {
```
imbhargav5, line 31: We don't need to use `useEffectOnce` I believe. We are already using `useUpdateEffect` so that is fine.

imbhargav5 (changes requested):
So the 3 hooks that we will use here are `useDidMount`, `useUpdateEffect` and `useWillUnmount` . Great work man!

src/__tests__/useLifecycleLogger.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+import React from 'react';
+import { useLifecycleLogger } from '../hooks/useLifecycleLogger';
+
+describe('useLifecycleLogger', () => {
+  it('should be defined', () => {
+    expect(useLifecycleLogger).toBeDefined();
+  });
+});
+
```
imbhargav5, line 12: Let's also add test cases using  `@testing-library/react-hooks`. You can check our other hooks for examples @iampika . Great work so far!

imbhargav5 (changes requested):

src/hooks/useLifecycleLogger.ts
```diff
@@ -0,0 +1,27 @@
...
+import { useWillUnmount } from './useWillUnmount';
+
+/**
+ * useLifecycleLogger hook
+ *
+ * @param componentName Name of the component
+ * @param rest
+ */
+const useLifecycleLogger = (componentName: string = 'Component', ...rest) => {
+  useEffectOnceWhen(() => {
```
imbhargav5, line 12: Just the one small change here. We need to use `useDidMount` here instead of `useEffectOnceWhen`

imbhargav5 (changes requested):

PR #487, opened 2021-05-15
feat: add useCookieState hook with tests

pikaatic:


src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,31 @@
...
+  string | null,
+  (newValue: string, options?: Cookies.CookieAttributes) => void,
+  () => void
+] => {
+  const [value, setValue] = useState<string | null>(
+    () => Cookies.get(cookieName) || null
+  );
+
+  const updateCookie = useCallback(
+    (newValue: string, options?: Cookies.CookieAttributes) => {
```
imbhargav5, line 16: The `options` here are best passed at `useCookie` level. 

```jsx
const [] = useCookie('my-cookie', options)
```

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,31 @@
+import Cookies from 'js-cookie';
+import { useCallback, useState } from 'react';
+
+const useCookieState = (
```
imbhargav5, line 4: JSDoc comments are missing. We need to add them here :D

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,31 @@
...
+  cookieName: string
+): [
+  string | null,
+  (newValue: string, options?: Cookies.CookieAttributes) => void,
+  () => void
+] => {
+  const [value, setValue] = useState<string | null>(
+    () => Cookies.get(cookieName) || null
```
imbhargav5, line 12: `Get` should also use the same options that were used in `set`

imbhargav5: Als this breaks on the server side. So on the server side we need to return null if window is not defined.

imbhargav5 (changes requested):
Also it would be great if we support this syntax as well. https://github.com/js-cookie/js-cookie#read

imbhargav5:
Looking good overall @iampika Great work!

src/hooks/useCookieState.ts
imbhargav5: So some comments are still missing. You can refer to the other hooks for guidance. Basically we need `useCookie` and also the description of what this hook does in there as well

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param cookieName Name of the cookie
+ * @returns {handler} A handler to interact with the cookie
+ */
+const useCookieState = (
+  cookieName: string,
+  cookieValue: string | null = null,
+  options?: Cookies.CookieAttributes
+): UseCookieStateType => {
+  const [value, setValue] = useState<string | null>(
+    () => Cookies.get(cookieName, options) || cookieValue
```
imbhargav5, line 21: We need to check for `window` here. so 

```
const [value, setValue] = useState<string | null>(
     () => {
   if(typeof window !== 'null'){
      return Cookies.get(cookieName, options) || cookieValue
   }else{
      return null;
   }
});
```

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+    [cookieName, options]
+  );
+
+  const deleteCookie = useCallback(() => {
+    Cookies.remove(cookieName);
+    setValue(null);
+  }, [cookieName]);
+
+  if (!window) return null;
```
imbhargav5, line 37: We don't need window check here.

imbhargav5 (changes requested):

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useCallback, useState } from 'react';
+
+type UseCookieStateType = {
+  value: string | null;
+  updateCookie: (newValue: string) => void;
+  deleteCookie: () => void;
+} | null;
+
+/**
+ *
```
imbhargav5, line 11: We need more comments here.

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,44 @@
...
+  options?: Cookies.CookieAttributes
+): UseCookieStateType => {
+  const [value, setValue] = useState<string | null>(() => {
+    if (typeof window !== undefined) {
+      return Cookies.get(cookieName, options) || cookieValue;
+    }
+
+    return null;
+  });
+
```
imbhargav5, line 27: We need useCookieState to listen for updates to this cookie from external events. So basically a listener on cookie change similar to how our useLocalstorageState works. WE need something for cookies .

imbhargav5 (changes requested):

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    [cookieName, options]
+  );
+
+  const deleteCookie = useCallback(() => {
+    Cookies.remove(cookieName);
+    setValue(null);
+  }, [cookieName]);
+
+  return { deleteCookie, updateCookie, value };
```
imbhargav5, line 41: `[value, setValue, removeValue]`

imbhargav5 (changes requested):

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,44 @@
...
+} | null;
+
+/**
+ *
+ * @param cookieName Name of the cookie
+ * @returns {handler} A handler to interact with the cookie
+ */
+const useCookieState = (
+  cookieName: string,
+  cookieValue: string | null = null,
```
imbhargav5, line 17: Rename to `key` & 'initialState'.

src/hooks/useCookieState.ts
```diff
@@ -0,0 +1,44 @@
...
+  options?: Cookies.CookieAttributes
+): UseCookieStateType => {
```
imbhargav5, line 19: 4th argument JSCookie arguments. At least have `converter` option.

imbhargav5 (changes requested):

qiweiii:
@imbhargav5 Seems ready to merge after sovling conflicts

PR #493, opened 2021-05-19
Add fix for SSR dev server hydration

allan2:
This solves the dev server hydration issue on Next.js.

Note: I have not tested on other SSR instances like Gatsby.

### DIFF
--- a/src/hooks/useWindowSize.ts
+++ b/src/hooks/useWindowSize.ts
@@ -31,13 +31,7 @@ function getDimensions(): WindowDimensions {
  * @returns Dimensions of the window
  */
 export function useWindowSize(): WindowDimensions {
-  const [windowSize, setWindowSize] = useState<WindowDimensions>(() => {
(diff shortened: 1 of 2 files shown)


src/hooks/useWindowSize.ts
```diff
@@ -31,13 +31,7 @@ function getDimensions(): WindowDimensions {
...
  */
 export function useWindowSize(): WindowDimensions {
-  const [windowSize, setWindowSize] = useState<WindowDimensions>(() => {
-    if (typeof window !== 'undefined') {
-      return getDimensions();
-    } else {
-      return nullDimensions;
-    }
-  });
+  const [windowSize, setWindowSize] = useState<WindowDimensions>(nullDimensions);
```
imbhargav5, line 34: This makes the window size null on first render on client side too, which is not right.

allan2: I think the useEffect sets it on mount. It's working in my example 5.

PR #502, opened 2021-06-18
Feat: useResizeObserver

AdrienLemaire:
Inspired by the existing [useMutationObserver](https://react-hooks.org/docs/useMutationObserver), I wrote a similar hook using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) and thought this might be a useful addition to the library.

PR #503, opened 2021-06-21
feat: useResizeObserver hook

AdrienLemaire:
Recreated the PR #502 using the main branch.

> Inspired by the existing [useMutationObserver](https://react-hooks.org/docs/useMutationObserver), I wrote a similar hook using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) and thought this might be a useful addition to the library.


imbhargav5:
Hey @AdrienLemaire the build seems to be failing. Perhaps some types broke.

AdrienLemaire:
@imbhargav5 please consider upgrading the typescript dependency version for rooks.
I cannot see ResizeObserver in 4.0.5, but my project using typescript 4.3.4 is correctly typed:

https://github.com/microsoft/TypeScript/blob/v4.3.4/lib/lib.dom.d.ts#L12515

imbhargav5:
Gotcha. Looks like I have a big weekend in front of me 😂

src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (size.height !== height || size.width !== width)
+        setSize(() => ({ height, width }));
+    });
+
+    if (ref.current) observer.observe(ref.current);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [ref.current]); // eslint-disable-line react-hooks/exhaustive-deps
```
imbhargav5, line 38: Using `Ref`.current with `useEffect` won't work. Instead we should use callback refs.

imbhargav5 (changes requested):

PR #506, opened 2021-07-06
fix: memo hooks

### LINKED ISSUE #500: Missing memoization
**Describe the bug**
Several hooks are missing useMemo or useCallback and returning new objects/function at every render. This cause issues in effects that have dependencies on these, triggering them when they shouldn't.

**To Reproduce**
Example using **useTimeout**

```tsx
  const myRef = useRef(null);
  const callback = useCallback(() => {
      myRef.current.x = x;
      myRef.current.y = y
    }, [])
  const { start: addInitialPositions, clear } = useTimeout(callback, 0);

  useEffect(() => {
    addInitialPositions();
    return clear;
  }, [addInitialPositions, clear]);
```

I want my effect to be executed only once, and using a timeout of 0 to change the priority of the callback in the execution stack.
But useTimeout returns new functions for start & clear at every render, which retriggers the useEffect every time.

**Expected behavior**
* start should be memoized
* clear should be memoized

in https://github.com/imbhargav5/rooks/blob/535f788bbd1196fed4ae88cfd2afea36f85a0c53/src/hooks/useTimeout.ts#L27-L37
```diff
-  function start() {
-    setIsTimeoutActive(true);
-  }
+ const start = useCallback(() => setIsTimeoutActive(true), [])
```

**Alternative solution**
Force an empty dependency list

```tsx
  useEffect(() => {
    addInitialPositions();
    return clear;
  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
Unfortunately, this hides the problem that functions are unnecessarily being recreated. Since our component is rerendered at a 60FPS rate, we have 60 new start & clear functions created at every second, increasing the number of garbage collector calls, which can create visual lags.

Please consider reviewing all your hooks to insure that objects & functions are only recreated when needed (when a dependency is changing), thanks.

phmngocnghia:
resolve #500 

**Address memo issues + add tests on these hooks:**
- useCounter
- useInput
- useLocalstorage
- useLocalstorageState
- useMapState
- useQueueState
- useSelect
- useSelectableList
- useSessionstorageState
- useStackState
- useTimeout
- useMutilSelectTableList
- useSelectTableList


imbhargav5:
cc @AdrienLemaire We are doing memoization work here. We should have this one finished pretty soon.

imbhargav5:
This seems to have a bunch of `'` to `"` changes. 😞  Makes it a bit hard to review

imbhargav5 (approved):

Issue #501, opened 2021-06-16
useThrottle example not working


robations:
**Describe the bug**
When typing in the input of the useThrottle example, the "throttle value" remains empty.

I tried adjusting the code example, but wasn't really sure what the expected values were or to find a fix.

Console warning:

```
Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.
p
div
App@https://t0qql.csb.app/src/App.js:26:41
```

**To Reproduce**
1. Go to https://react-hooks.org/docs/usethrottle/#codesandbox-example
2. Type in input "abcdef"
3. Actual value updates, but throttle value doesn't
4. Warning (above) in console

**Expected behavior**

Guessing the throttle value should echo the actual value after a 1 second delay.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Firefox
 - Version 89

Also reproduced issue with Chrome (Mac).

imbhargav5:
Hi @robations Apologies for the delay. Was a bit caught up with work. I will get to them this weekend. Thank you!

robations:
No problem, I know the feeling.

imbhargav5:
Fix is WIP. It will be out in a couple of days. Thank you

qiweiii:
@imbhargav5 I have looked at the example, the hook should be working fine, but the sandbox example is not using useThrottle correctly. My updated example is below:

```js
import React, { useState, useRef } from "react";
import { useThrottle } from "rooks";

function App() {
  const [text, setText] = useState("");
  const [throttleValue, setThrottleValue] = useState("");
  const [throttledFn, isReady] = useThrottle(changeThrottleValue, 1000);
  const ref = useRef(null);

  function changeThrottleValue() {
    setThrottleValue(ref.current.value);
  }

  return (
    <div>
      <h1>Rooks : useThrottle example</h1>
      <input
        ref={ref}
        onChange={(e) => {
          setText(e.target.value);
          if (isReady) throttledFn();
        }}
      />
      <p>Actual value: {text}</p>
      <p>
        Throttled value (being updated at most once per second): {throttleValue}
      </p>
    </div>
  );
}

export default App;
```

Please try it out

PR #514, opened 2021-07-24
docs(useThrottle): update useThrottle example and sandbox

qiweiii:
#501

### DIFF
--- a/docs/useThrottle.md
+++ b/docs/useThrottle.md
@@ -78,9 +78,9 @@ render(<Demo />);
   src="https://codesandbox.io/embed/usethrottle-forked-0nf94?fontsize=14&hidenavigation=1&theme=dark"
   style={{
     width: "100%",
-    height: "500px",
+    height: 500,
     border: 0,
-    border-radius: "4px",
+    borderRadius: 4,
     overflow: "hidden"
   }}
   title="useThrottle"
--- a/docs/useThrottle.md
+++ b/docs/useThrottle.md
@@ -20,48 +20,72 @@ import { useThrottle } from 'rooks';
 
 ## Usage
 
-```jss
-function ThrottleDemo() {
-  const [number, setNumber] = useState(0);
-  const addNumber = () => setNumber(number + 1);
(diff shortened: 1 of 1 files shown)


docs/useThrottle.md
```diff
@@ -20,48 +20,72 @@ import { useThrottle } from 'rooks';
...
-      <button onClick={addNumber}>Add number</button>
-      <button onClick={addNumberThrottled}>Add number throttled</button>
-    </>
+    <div>
+      <h1>Rooks : useThrottle example</h1>
+      <input
+        ref={ref}
+        onChange={(e) => {
+          setText(e.target.value);
+          if (isReady) throttledFn();
```
imbhargav5, line 45: `isReady` is just an indication for the user. But the conditional is not needed here.

docs/useThrottle.md
```diff
@@ -20,48 +20,72 @@ import { useThrottle } from 'rooks';
...
+function Demo() {
+  const [text, setText] = useState("");
+  const [throttleValue, setThrottleValue] = useState("");
+  const [throttledFn, isReady] = useThrottle(changeThrottleValue, 1000);
+  // isReady is a boolean that tells you whether calling throttledFn at that point
   // will fire or not.
   // Once the timeout of 1000ms finishes, isReady will become true to indicate that the next time
-  // addNumberThrottled is called it will run right away.
+  // throttledFn is called it will run right away.
+  const ref = useRef(null);
```
imbhargav5, line 32: What does this `ref` do? 🤔

imbhargav5 (changes requested):

docs/useThrottle.md
qiweiii: Got it

docs/useThrottle.md
qiweiii: It was from previous code, I think we don't need to use a ref here, I will remove it

imbhargav5 (approved):

imbhargav5 (approved):

PR #496, opened 2021-05-25
feat: add useSet hook

pikaatic:

PR #516, opened 2021-07-24
Fix useSet

qiweiii:
Continue from #496


imbhargav5:
This PR originally by @iampika was an 80% direct rip-off from react-use. Let's recreate this and try to solve this in our own way @qiweiii .

PR #523, opened 2021-07-26
feat: useResizeObserver hook updated

qiweiii:
Continue from #503


src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,37 @@
...
+ * useResizeObserver hook
+ *
+ * Returns a resize observer for a React Ref and fires a callback
+ * https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
+ *
+ * @param {RefObject<HTMLElement | null>} ref React ref on which resizes are to be observed
+ * @param {ResizeObserverCallback} callback Function that needs to be fired on resize
+ * @param {Options} options An options object allowing you to set options for the observation
+ */
+function useResizeObserver(
```
imbhargav5, line 18: I think I want to get away from the pattern of using `mutableRef`s here and instead use `Callback` refs. Also let's change this hook to be `useResizeObserverRef` and create a ref in the hook and then expose it out. 

```jsx
const [ref] = useResizeObserver(callback, options)

<div ref={ref}>

</div>
```

imbhargav5 (changes requested):

src/hooks/useResizeObserver.ts
imbhargav5: Main reason for it being that `mutableRefs` don't really track node changes. Apologies for not mentioning this in advance but this is a better solution in the long run and I would appreciate this change a lot! @qiweiii

src/hooks/useResizeObserverRef.ts
```diff
@@ -0,0 +1,49 @@
+import { useCallback, useEffect, useState } from "react";
+import type { CallbackRef, HTMLElementOrNull } from "../utils/utils";
+
+interface ResizeObserverOptions {
```
imbhargav5, line 4: Instead of using types from here, are there types from ResizeObserver natively somewhere? Just a thought.

src/hooks/useResizeObserverRef.ts
```diff
@@ -0,0 +1,49 @@
...
+      const observer = new ResizeObserver(callback);
+
+      // Start observing the target node for resizes
+      observer.observe(node, options);
+
+      return () => {
+        observer.disconnect();
+      };
+    }
+  }, [node, callback, options]);
```
imbhargav5, line 40: We will need to use `useFreshTick` here, which will keep the callback fresh. https://react-hooks.org/docs/useFreshTick

Let's also add tests for this one. Excellent work so far! @qiweiii

imbhargav5 (changes requested):

src/__tests__/useResizeObserverRef.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+      );
+    };
+  });
+
+  afterEach(cleanup);
+
+  it("should be defined", () => {
+    expect(useResizeObserverRef).toBeDefined();
+  });
+});
```
qiweiii, line 79: Since jsDOM not supporting layout code like ResizeObserver or MutationObserver, so I cannot test behaviour of this hook

imbhargav5:
Your commits were released in another PR. Thanks @qiweiii !

PR #526, opened 2021-07-27
Make console.warn in useMediaMatch opt-in

ackl:
This hook currently clutters logs when used in SSR projects.


imbhargav5:
@ackl This is a great idea. In fact I will add this to more utils. Instead of an argument, let's try using `process.env.NODE_ENV` and use that instead. This will mean that we won't have this warning on the server side in production.

imbhargav5:
So basically we can try
```jsx
if (process.env.NODE_ENV==='development') {
  console.warn('useMediaMatch cannot function as window is undefined.')
};
```

ackl:
Dev server logs also being free of this clutter is actually the main motivation behind this PR 😅

What about reading a env var named `LOG_LEVEL` or something instead?

ackl:
@imbhargav5 I've created a logger module which wraps console methods which some checks for env vars, stdout/err should only have output now if `NODE_ENV!=='production' && LOG_LEVEL==='debug'` or `NODE_ENV==='production' && LOG_LEVEL === 'production'`

pls lemme know ur thoughts

imbhargav5:
Wow @ackl This is a pretty impressive PR. I like the `logger` module. 

I see some  extra `prettier` changes though which is making this hard to review.

src/utils/logger.ts
```diff
@@ -0,0 +1,27 @@
+const methods = ["log", "info", "warn", "error", "debug", "exception"] as const;
+type Methods = typeof methods[number];
+
+function consoleWrap(method: Methods) {
+  return function (...args) {
+    if (process.env.NODE_ENV !== "production") {
+      if (process.env.LOG_LEVEL === "debug") {
```
imbhargav5, line 7: I personally think that `NODE_ENV` is enough for now and I don't want to complicate things with `LOG_LEVEL` at this point. 

We will log on development and not in production. That should keep super simple.

imbhargav5: This is mainly because packages like react and styled-components have adopted this convention of only checking for NODE_ENV which I think is both sturdy and also simple at the same time. I would prefer to keep it that way.

imbhargav5 (changes requested):

imbhargav5:
I think I want to make a few changes to this one. I will try combining `logger` with `warning` and see what we can do something this week.

tk-alow:
thanks for picking this up. apologies for all the `prettier` noise, you guys may want to think about setting a more explicit prettierrc )))

ackl:
thanks for picking this up. apologies for all the `prettier` noise, you guys may want to think about setting a more explicit prettierrc )))

imbhargav5:
Haha. It's getting hard to review because of the prettier stuff. @qiweiii . I will take some time for this one to review. Feedback should be out tomorrow.

imbhargav5:
We need to figure out how to get this merged. Let's try this @qiweiii 
Let's move the logger into a different PR. 
1. Ideally we should have the following utils in usage
    a. invariant
    b. warning (That only warns when not in production)
2. Then use the warning here.

qiweiii:
@imbhargav5 what do you mean by `invariant`

Issue #530, opened 2021-07-30
useOutsideClickRef doesn't detect node changes


AdrienLemaire:
**Describe the bug**
When the node ref used by useOutsideClickRef re-renders or if its children change, clicking within a new element in that node will trigger useOutsideClickRef's callback.

**To Reproduce**
1. Open a menu used as ref by useOutsideClickRef
2. click on an icon in the menu that will replace said icon with a new one (in my case a button to toggle on/off sound)
3. click on the new icon => useOutsideClickRef callback triggered (in my case, incorrectly closes the menu).

**Expected behavior**
useOutsideClickRef memoizedCallback should rebuild the callback when the node is replaced or mutated.
Maybe using useMutationObserverRef would be the solution.

imbhargav5:
@AdrienLemaire Will do this on top priority. Could you help us with a code sandbox? That would be super useful.

AdrienLemaire:
@imbhargav5 as a temporary solution, I'm returning both icons, and toggling the hidden css style, so this is not urgent to me.

Sorry, I've finished my week and have to go. Tried to do a sandbox here, but the result is a bit different  (clicking on the button at any time will throw the alert)
https://codesandbox.io/s/useoutsideclickref-forked-m4ztt?file=/src/App.js

imbhargav5:
@qiweiii can you check this out?

qiweiii:
@imbhargav5 Got it, I'll take a look at it soon

imbhargav5:
can you check if the latest version fixes your issue @AdrienLemaire ?

imbhargav5:
Closing this for now but feel free to reopen.

PR #534, opened 2021-07-31
fix(useOutsideClickRef): fix child update logic and event listener phase

qiweiii:
issue #530 

The main cause is that the callback is called in event bubbling phase which happens after the click event for icon change

### DIFF
--- a/src/hooks/useOutsideClickRef.ts
+++ b/src/hooks/useOutsideClickRef.ts
@@ -4,8 +4,8 @@ import {
   useRef,
   useCallback,
   useState,
-} from 'react';
-import type { HTMLElementOrNull, CallbackRef } from '../utils/utils';
+} from "react";
+import type { HTMLElementOrNull, CallbackRef } from "../utils/utils";
 
 /**
  * useOutsideClickRef hook
@@ -25,9 +25,10 @@ function useOutsideClickRef(
 
   const memoizedCallback = useCallback(
     (e: MouseEvent) => {
-      if (node && !node.contains(e.target as Element)) {
-        savedHandler.current(e);
+      if (!node || node.contains(e.target as Element)) {
+        return;
       }
+      savedHandler.current(e);
     },
     [node]
   );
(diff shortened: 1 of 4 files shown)


imbhargav5:
@qiweiii Looks great!

src/hooks/useOutsideClickRef.ts
```diff
@@ -25,9 +25,10 @@ function useOutsideClickRef(
 
   const memoizedCallback = useCallback(
     (e: MouseEvent) => {
-      if (node && !node.contains(e.target as Element)) {
-        savedHandler.current(e);
+      if (!node || node.contains(e.target as Element)) {
```
imbhargav5, line 28: Does this fix the issue? 🤔

qiweiii: No it does not, I realised it's the same logic as before

imbhargav5: Yeah we need to figure out what the cause of the bug is. Perhaps `currentTarget` might be a solution?

qiweiii: Yes, the cause of the bug is that we did not use event capturing.

When user click a descendent element inside the ref, the default behaviour was that the `e.target` node's click event handler being called first, and if this changed the child element inside, the child will be updated immediately. After this, the click event propagates up, until document's click event handler being called, at that time, the `node` saved in our hook  is not up-to-date, the child element inside it is the old one, so in our `memoizedCallback`, `!node.contains(e.target as Element)` will be `true`, that's the cause of the bug.

So I changed to event capturing by adding `true` to `document.addEventListener("click", memoizedCallback, true);`, this will trigger the `document`'s event handler first, then does down to the child elements 

[example](https://codesandbox.io/s/useoutsideclickref-forked-45qbn?file=/src/useOutClick.ts)

imbhargav5: Makes total sense @qiweiii . I think we should update `useOutsideClick` hook too in the same way.

qiweiii: Shall I create another PR, or just add it here?

imbhargav5 (approved):

imbhargav5 (approved):

Issue #545, opened 2021-08-01
docs: sidebar disappears when click on Documentation


qiweiii:
**Describe the bug**
Asidebar disappears when click on Documentation

**To Reproduce**
1. Go to https://react-hooks.org/
2. Click on 'Documentation' at the top

**Expected behavior**
Sidebar appears on the left

**Screenshots**
![image](https://user-images.githubusercontent.com/32790369/127778444-156fd505-ba67-4555-896d-f3ed1993d0ac.png)

**Desktop (please complete the following information):**
 - Browser [chrome, safari]

PR #547, opened 2021-08-02
docs: fix missing sidebar when click on docmentation

qiweiii:
#545

The cause is that we removed 'rooks' page from sidebar, it corresponds to '/docs' route. Clicking on 'Documentation' brings use to '/docs', so i added it back to sidebar to fis it.

If we don't want rooks page at all. We need to change some other page to be '/docs'

### DIFF
--- a/sidebars.json
+++ b/sidebars.json
@@ -3,10 +3,7 @@
     {
       "label": "Quick Start",
       "collapsed": false,
-      "items": [
-        "getting-started",
-        "motivation"
-      ],
+      "items": ["getting-started", "motivation", "rooks"],
       "type": "category"
     },
     {
--- a/docs/rooks.md
+++ b/docs/rooks.md
@@ -5,86 +5,12 @@ sidebar_label: rooks
 slug: /
 ---
 
-<br/>
-<br/>
+
 <img src="https://github.com/imbhargav5/rooks/raw/master/.github/assets/logo-dark.png" height="auto" width="100%" />
 
 <br/>
-<br/>
-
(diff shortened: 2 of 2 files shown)


imbhargav5:
> The cause is that we removed 'rooks' page from sidebar, it corresponds to '/docs' route. Clicking on 'Documentation' brings use to '/docs', so i added it back to sidebar to fis it.

> If we don't want rooks page at all. We need to change some other page to be '/docs'
@qiweiii Yeah lets do that.

qiweiii:
@imbhargav5 Which page do you want '/docs' routes to? I was thinking abt 'getting-started' page, but then both 'Getting started' and 'Documentation' will routes to 'getting'started'.

imbhargav5:
@qiweiii Let's remove the unnecessary links and keep it simple. Good call!

sidebars.json
```diff
@@ -3,10 +3,7 @@
     {
       "label": "Quick Start",
       "collapsed": false,
-      "items": [
-        "getting-started",
-        "motivation"
-      ],
+      "items": ["getting-started", "motivation", "rooks"],
```
imbhargav5, line 6: I think rooks can be removed here. @qiweiii

imbhargav5 (changes requested):

sidebars.json
qiweiii: If we remove it, the sidebar will disappear when to go 'docs/' page.

imbhargav5 (approved):

PR #552, opened 2021-08-11
feat(useDebouncedValue): add useDebouncedValue hook

maciekgrzybek:
WIP - just hook + tests, without example, mds etc as I want to make sure I understood correctly what this hook is supposed to be doing :)

### DIFF
--- a/src/__tests__/useDebouncedValue.spec.ts
+++ b/src/__tests__/useDebouncedValue.spec.ts
@@ -4,50 +4,62 @@
 import { act, renderHook } from "@testing-library/react-hooks";
 import { useDebouncedValue } from "../hooks/useDebouncedValue";
 
-jest.useFakeTimers();
+jest.useFakeTimers('modern');
 
 describe("useDebouncedValue", () => {
   it("should be defined", () => {
     expect(useDebouncedValue).toBeDefined();
   });
 
-  it("should return null if the timeout has not been reached", () => {
+  it("should initialize with first value if timeout is not reached and initializeWithNull is false", () => {
     const mockValue = "mock_value";
 
     const { result } = renderHook(() => useDebouncedValue(mockValue, 200));
 
     act(() => {
-      expect(result.current).toBeNull();
+      expect(result.current[0]).toBe(mockValue);
     });
   });
 
-  it("should returns updated value if the timeout has been reached", () => {
+
+  it("should return null if the timeout has not been reached and initializeWithNull is true", () => {
     const mockValue = "mock_value";
 
-    const { result } = renderHook(() => useDebouncedValue(mockValue, 200));
+    const { result } = renderHook(() => useDebouncedValue(mockValue, 200, {initializeWithNull : true}));
 
     act(() => {
+      expect(result.current[0]).toBeNull();
+    });
+  });
+
+  it("should returns updated value if the timeout has been reached and initializeWithNull is true", () => {
+    const mockValue = "mock_value";
+
+    const { result } = renderHook(() => useDebouncedValue(mockValue, 200, {initializeWithNull : true}));    
+    act(() => {
+      expect(result.current[0]).toBeNull();
       jest.runAllTimers();
-      expect(result.current).toBe(mockValue);
+      expect(result.current[0]).toBe(mockValue);
     });
   });
 
-  it("should respect the timeout value", () => {
+  it("should respect the timeout value if initializedWithNull is true", () => {
     const mockValue = "mock_value";
     const mockTimeout = 200;
 
     const { result } = renderHook(() =>
-      useDebouncedValue(mockValue, mockTimeout)
+      useDebouncedValue(mockValue, mockTimeout, {initializeWithNull : true})
     );
 
     act(() => {
-      jest.advanceTimersByTime(mockTimeout - 1);
-      expect(result.current).toBe(null);
+      jest.advanceTimersByTime(mockTimeout - 5);
+      expect(result.current[0]).toBe(null);
     });
 
     act(() => {
-      jest.advanceTimersByTime(1);
-      expect(result.current).toBe(mockValue);
+      // accounting for timing issues as it's not exactly accurate
+      jest.advanceTimersByTime(10);
+      expect(result.current[0]).toBe(mockValue);
     });
   });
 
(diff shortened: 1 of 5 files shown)


src/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from "react";
+
+function useDebouncedValue<T = unknown>(value: T, timeout: number) {
+  const [updatedValue, setUpdatedValue] = useState<T | null>(null);
```
imbhargav5, line 4: It might be worth to use `lodash.debounce` here to be super sure of the behaviour. We are already using it in `useDebounce`

imbhargav5 (changes requested):

src/hooks/useDebouncedValue.ts
maciekgrzybek: Not sure how that's supposed to work with debounce to be fair... I changed the implementation to this:
```
function useDebouncedValue<T = unknown>(value: T, timeout: number) {
  const [updatedValue, setUpdatedValue] = useState<T | null>(null);
  useEffect(() => {
    debounce(() => setUpdatedValue(value), timeout);
  }, [value, timeout]);

  return updatedValue;
}
```

But my tests are failing.

maciekgrzybek: Also just wondering... do we really need `lodash` here? I mean, this hook is pretty straightforward and I think that `setTimeout` would be enough?

qiweiii: For correct debounce behaviour, it will cancel the previous timer everytime when the function is called. So we need to change the code for debounce behaviour or use lodash debounce or maybe `useDebounce` hook.

I think we want something similar to the usage of React `useDeferredValue`, refer to [this section in react doc](https://reactjs.org/docs/concurrent-mode-patterns.html#deferring-a-value). We may also need to expose a debounced `setUpdatedValue` function (not so sure abt this)

maciekgrzybek: I might be missing something here, but isn't `clearTimeout` doing exactly that? Also about exposing the function, I don't think we should. This hooks supposed to expose only the value in my opinion.

imbhargav5: Set myself a reminder to comeback to this in the weekend. Will get it shipped in one form or another tomorrow @maciekgrzybek @qiweiii . Thanks for your hardwork!

imbhargav5: What do you think of this? @maciekgrzybek @qiweiii ? Just checking if this solves the problem and to look for the best way of doing this. 👍🏽 

```jsx

function useDebouncedValue<T = unknown>(value: T, timeout: number) {
  const [updatedValue, setUpdatedValue] = useState<T | null>(null);
  const debouncedSetUpdatedValue = useDebounce(setUpdatedValue);
  useEffect(() => {
       debouncedSetUpdatedValue(updatedValue);
  }, [updatedValue])
 // No need to add `debouncedSetUpdatedValue ` to dependencies as it is a ref.current.

  return updatedValue;

}


```

imbhargav5: I don't want to overcomplicate it but at the same time I think `setUpdatedValue` might be useful here.

maciekgrzybek: This is not working unfortunately, the tests are failing. I tried something similar before - same result.

qiweiii: I also tried something similar but tests are failing. Found this [issue](https://github.com/facebook/jest/issues/3465) about jest.runAllTimers may be related

src/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,18 @@
...
+  timeout: number
+): [T | null, React.Dispatch<React.SetStateAction<T | null>>, Function] {
+  const [updatedValue, setUpdatedValue] = useState<T | null>(null);
+  const debouncedSetUpdatedValue = useDebounce(setUpdatedValue, timeout);
+  useEffect(() => {
+    debouncedSetUpdatedValue(value);
+  }, [value]);
+  // No need to add `debouncedSetUpdatedValue` to dependencies as it is a ref.current.
+
+  return [updatedValue, setUpdatedValue, debouncedSetUpdatedValue];
```
qiweiii, line 15: Returning both `setUpdatedValue` and `debouncedSetUpdatedValue` here, because if user is using debounced value already, I think they may also want future updates to be debounced.

Btw, tests are passing now.

src/__tests__/useDebouncedValue.spec.ts
```diff
@@ -4,7 +4,7 @@
 import { act, renderHook } from "@testing-library/react-hooks";
 import { useDebouncedValue } from "../hooks/useDebouncedValue";
 
-jest.useFakeTimers();
+jest.useFakeTimers("modern");
```
imbhargav5, line 7: Lol. I did the exact same thing and was about to push.

src/hooks/useDebouncedValue.ts
imbhargav5: Not sure if we should expose this right now. `debouncedSetUpdatedValue`. Let's keep it simple for now and add it if sufficient demand is there. Besides they can already do a `debouncedSetUpdatedValue` by simply updating the `value`.

imbhargav5 (changes requested):

imbhargav5:
@qiweiii I think want to make a few changes in the behavior. `initializingWithNull` should probably be an option and not the default behavior. Mainly because this hook will probably be used with lots of inputs and they might want the input to be initialised with a string and they would want the debounced value to also be the same initialized value in the first render.

imbhargav5:
Force pushed @qiweiii to overwrite your changes mainly because I think we started with the same intention of fixing tests but I had a few things more to add. Please review it @qiweiii @maciekgrzybek and let me know your thoughts.

src/__tests__/useDebouncedValue.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    );
+
+    act(() => {
+      jest.advanceTimersByTime(mockTimeout - 5);
+      expect(result.current[0]).toBe(null);
+    });
+
+    act(() => {
+      // accounting for timing issues as it's not exactly accurate
+      jest.advanceTimersByTime(10);
```
imbhargav5, line 61: I couldn't figure this out. Trying it with `mockTimeout - 1` and  `1` didn't help me. Can you take a look here @qiweiii because you had already fixed it on your end in the previous `HEAD`

imbhargav5: I think it might be because of a slight lag in `useEffect` not entirely sure. But `mockTimeout - x` and `x+1` work but not   `mockTimeout - x` and `x` 🤔

qiweiii: @imbhargav5 move `expect` out of `act` will work

PR #553, opened 2021-08-12
Add eager possibility for useIntervalWhen

mar1u50:
When using useIntervalWhen there is no possibility to say: invoke the callback right before the interval elapses.
If you use useIntervalWhen for polling a service from the backend this is a very common scenario.

### DIFF
--- a/src/hooks/useIntervalWhen.ts
+++ b/src/hooks/useIntervalWhen.ts
@@ -1,4 +1,4 @@
-import { useRef, useEffect } from 'react';
+import { useRef, useEffect } from "react";
 
 /**
  * A setInterval hook that calls a callback after a interval duration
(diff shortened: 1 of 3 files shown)


src/hooks/useIntervalWhen.ts
```diff
@@ -25,6 +27,9 @@ function useIntervalWhen(
 
   useEffect(() => {
     if (when) {
+      if (eager) {
+        callback();
```
imbhargav5, line 31: Won't this run even before the timeout? This will run well before the interval. Am I missing something?

imbhargav5 (changes requested):

src/hooks/useIntervalWhen.ts
imbhargav5: What do you think of this @qiweiii ?

qiweiii: @imbhargav5 Yes, it runs the callback before timer starts.

I think @mar1u50 means when we start polling a backend service, we usually don't want to wait for **the first interval**, we want to send out the first request right away, then start the timer and wait for a while and send another request and repeat. I think it is a common scenario. Good point @mar1u50!

imbhargav5: Ah. Sounds good. I like the use case. Let's document this option in the Readme, add an example and finish this off @qiweiii

imbhargav5 (approved):

PR #555, opened 2021-08-14
chore(contributors): automate readme contributors list update

qiweiii:
- removed `contributors.yml` workflow since it does not update all contributors
- use the same tool to update contributors list on both readme and rooks website


imbhargav5:
Excellent work @qiweiii 🔥🔥🔥

PR #567, opened 2021-08-29
feat(useoutsideclick): detect click on outside iframe

phmngocnghia:
- Tested on: Safari, FF, Chrome
- Detect click outside on different iframe element
- Adds detectIframe config flag that defaults to `true`, can be turned off if users want to detect by themself because the logic is not perfect.
- Got the idea from https://stackoverflow.com/a/23231136


imbhargav5:
Hey @phmngocnghia Thanks for this one! Would you like to also add a code sandbox example showing this usage? That will help me to review and also others to understand the usage.

phmngocnghia:
@imbhargav5 Sure!

I pushed my example to this repository using ESM version of the hook generated from built: https://github.com/phmngocnghia/demo-rooks

The idea is the click outside hook should be able to detect if my component is in the main window, and I click into an iframe, it should emit the callback... E.g: This "specialized" hook: https://github.com/wellyshen/react-cool-onclickoutside#detecting-iframe-clicks

imbhargav5:
@phmngocnghia Ah I see. I think we should keep this consistent and put them in the docs examples. What do you think?

phmngocnghia:
@imbhargav5 I can update the new sandbox example. Do we have a pre-release version of the SDK? I'm not sure how do I install the version that contains changes on my PR on code sandbox.

imbhargav5:
@phmngocnghia unfortunately we don't. For now you can just copy paste the hook into the sandbox and use it. 😞

phmngocnghia:
@imbhargav5 I have added an example. What do you think about a GitHub action that can publish a pre-release version for each PR?

imbhargav5:
@phmngocnghia For each PR? I like the idea and I think it will solve a purpose. But I feel it will make the build a lot slower. Perhaps we can create a release on a command. Like /create-npm-release-from-pr. What do you think ? @qiweiii @simbathesailor @maciekgrzybek

versioned_docs/version-5.6.0/useOutsideClick.md
```diff
@@ -41,18 +48,12 @@ render(<Demo />);
...
-    border: 0,
-    borderRadius: 4,
-    overflow: "hidden"
-  }} 
-title="useNavigatorLanguage"
-allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
-sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
-/ >
+<iframe src="https://codesandbox.io/embed/useoutsideclick-ctkk6?expanddevtools=1&fontsize=14&hidenavigation=1&module=%2Fsrc%2FApp.tsx&theme=dark"
+     style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
```
imbhargav5, line 52: This needs to be in JSX. Not in HTML. So that means we need style={{}}

src/hooks/useOutsideClick.ts
```diff
@@ -8,32 +8,48 @@ import { useEffect, useRef, useCallback } from "react";
...
   }, []);
 
-  useEffect(() => {
-    savedHandler.current = handler;
-  });
+  const memoizedOnBlurCallback = useCallback((event: FocusEvent) => {
+    // Note: on firefox clicking on iframe triggers blur, but only on
+    // next event loop it becomes document.activeElement
+    // https://stackoverflow.com/q/2381336#comment61192398_23231136
+    setTimeout(() => {
```
imbhargav5, line 35: Can we use `useTimeout` here?

src/hooks/useOutsideClick.ts
```diff
@@ -8,32 +8,48 @@ import { useEffect, useRef, useCallback } from "react";
...
+  when: boolean = true,
+  detectIframe: boolean = true
 ): void {
   const savedHandler = useRef(handler);
 
-  const memoizedCallback = useCallback((e: MouseEvent) => {
-    if (ref && ref.current && !ref.current.contains(e.target as Element)) {
-      savedHandler.current(e);
+  const memoizedCallback = useCallback((event: MouseEvent) => {
+    if (ref?.current && !ref.current.contains(event.target as Element)) {
```
imbhargav5, line 26: Nice 👍🏽

imbhargav5 (changes requested):

imbhargav5:
@qiweiii Can you finish this PR? Let's get this merged tomorrow.

src/hooks/useOutsideClick.ts
qiweiii: using `useTimeout` here gives this error:
```
Error: Uncaught [Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
```

imbhargav5: Could you paste your snippet here?  @qiweiii

qiweiii: @imbhargav5 I found out Iwas using timeout incorectly previously, but still `useTimeout` is not suitable here because I have to declare `useTimeout`'s callback outside of this useCallback, and as `useTimeout` does not take aruguments, I cannot pass `event` into the timeout callback

PR #575, opened 2021-09-04
fix: better typescript types

### LINKED ISSUE #568: Better Typescript : Use native types than creating a new one
https://github.com/imbhargav5/rooks/blob/85168bafc72033d3fcb8905959a49fcdbaaf2484/src/hooks/useWindowSize.ts#L4

qiweiii:
Closes #568
@imbhargav5 For useGeolocation and useFullscreen, there is no native types available that match what these hooks require


imbhargav5:
@qiweiii Going to close this one as I believe there are changes to the hook internals to make. Please take a look at #578 PR. Let's do `useGeolocation` fixes in that PR and create a separate one for `useFullscreen`

Issue #577, opened 2021-09-04
Rewrite useGeoLocation hook


imbhargav5:
At this point we are using custom interfaces but I think we should come up with a better approach. Let's just return the geolocation object directly and also the error.

imbhargav5:
@qiweiii I am almost done with hook itself. Can you work on adding the test cases for it? We will need to mock geolocation in jest and get it worked. This is a breaking change.

imbhargav5:
@qiweiii The hook internals are done. Let's add tests to this one.

PR #578, opened 2021-09-04
fix(usegeolocation): rewrites geolocation internals

imbhargav5:
Fixes geolocation internals. Uses AbortController for unmount cleanup. Removes unnecessary types.

BREAKING CHANGE: Geolocation now returns the native geolocation position and error (if exists) in a
tuple.

re #577

### DIFF
--- a/docs/useGeolocation.md
+++ b/docs/useGeolocation.md
@@ -24,7 +24,7 @@ import { useGeolocation } from "rooks";
 
 ```jsx
 function App() {
-  const geoObj = useGeolocation();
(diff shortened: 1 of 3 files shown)


imbhargav5:
@qiweiii Let's add tests to this one and keep this in the shelf along with some other breaking changes. I think we have 2 more hooks that I would like fixed in a single `major` release.

qiweiii (approved):

imbhargav5:
Moving it to draft status so as to not accidentally merge it. We still need to fix tests.

PR #591, opened 2021-09-13
fix(useMediaMatch): safari 13.1 addListener fix

krijoh92:
closes 590

### DIFF
--- a/src/hooks/useMediaMatch.ts
+++ b/src/hooks/useMediaMatch.ts
@@ -25,9 +25,16 @@ function useMediaMatch(query: string): boolean {
     setMatches(matchMedia.matches);
     const listener = (event_: MediaQueryListEventMap['change']) =>
       setMatches(event_.matches);
-    matchMedia.addEventListener('change', listener);
 
-    return () => matchMedia.removeEventListener('change', listener);
+    if (matchMedia.addEventListener) {
+      matchMedia.addEventListener('change', listener)
+
+      return () => matchMedia.removeEventListener('change', listener)
+    } else {
+      matchMedia.addListener(listener)
+
+      return () => matchMedia.removeListener(listener)
+    }
   }, [matchMedia]);
 
   return matches;


src/hooks/useMediaMatch.ts
```diff
@@ -25,9 +25,16 @@ function useMediaMatch(query: string): boolean {
     setMatches(matchMedia.matches);
     const listener = (event_: MediaQueryListEventMap['change']) =>
       setMatches(event_.matches);
-    matchMedia.addEventListener('change', listener);
 
-    return () => matchMedia.removeEventListener('change', listener);
+    if (matchMedia.addEventListener) {
+      matchMedia.addEventListener('change', listener)
```
imbhargav5, line 30: Thanks for this! @krijoh92 

I just have one small suggestion just for consistency in code style that we already have. What do you think of rewriting it to look like this?


```jsx

matchMedia.addEventListener?.('change', listener)
matchMedia.addListener?.(listener)

return () => {
  matchMedia.removeEventListener?.('change', listener) 
  matchMedia.removeListener?.(listener)
}
```

krijoh92: I haven't tested yet, but wouldn't this make it trigger twice for all browsers that support `addEventListener` and `addListener`. `addListener` is only deprecated, but not removed in modern browsers

imbhargav5: Ah. I see. Agreed.

imbhargav5:
✅

reflog:
Hi @imbhargav5 , I see that this change was reverted in rooks master. Could you provide any details on the reasoning? Thanks in advance!

imbhargav5:
@reflog I need to look into this. Could be a mistake though. Let me get back tomorrow!

imbhargav5:
@reflog This seems like it was removed accidentally. But now that addEventListener method is present predominantly in all browsers, I am wondering if it's worth adding this back.

reflog:
@imbhargav5 I get it :) We still have clients on Safari 13.x and we'll have to downgrade rooks to 5.x 
Please let me know if there are no plans for supporting old Safari.
Thanks again for prompt reply!

imbhargav5:
@reflog I will be back from a vacation next week and I can make a PR then. First week of November.but if you are in a hurry, you can either make a PR or have do with rooksv5 until then.

imbhargav5:
@reflog Thanks for your help and understanding!  🙌

PR #595, opened 2021-09-18
chaore: add warning message for deprecated hooks

qiweiii:


docs/usePreviousImmediate.md
```diff
@@ -4,6 +4,12 @@ title: usePreviousImmediate
 sidebar_label: usePreviousImmediate
 ---
 
+:::warning
+
+`usePreviousImmediate` is deprecated, it will be removed in rooks v7. Please use [usePrevious](./usePrevious) instead.
```
imbhargav5, line 9: Actually, we should deprecate `usePrevious` which is buggy.

docs/useSessionstorageState.md
```diff
@@ -4,6 +4,12 @@ title: useSessionstorageState
 sidebar_label: useSessionstorageState
 ---
 
+:::warning
+
+`useSessionstorageState` is deprecated, it will be removed in rooks v7. Please use [useSessionstorage](/useSessionstorage) instead.
```
imbhargav5, line 9: I think I didn't explain this well. We should be deprecate `useSessionstorage`. It's buggy. Apologies. 😓

src/hooks/usePreviousImmediate.ts
```diff
@@ -7,6 +8,10 @@ import { useRef, useEffect } from 'react';
  * @returns The previous value
  */
 function usePreviousImmediate<T>(currentValue: T): T | null {
+  warning(
```
imbhargav5, line 11: This will print the warning on each render. 🤔 I would like the warnings to only be logged on the first render. Let's make a `useWarningOnMountInDevelopment` hook. It's a verbose name but it will be super clear. This hook will be for internal purposes and will print the warning only once for each hook. What do you think?

imbhargav5 (changes requested):

imbhargav5:
Looks terrific on first glance. Thanks @qiweiii . I will do a detailed review tomorrow!

PR #603, opened 2021-10-11
fix(useDebounce): return type of callback

krijoh92:
this way we the debounced function type is the same as the type of the callback we send in to `useDebounce`.


imbhargav5:
Me or @qiweiii  will handle the build error @krijoh92 🔥

qiweiii:
@imbhargav5 build error is fixed!

imbhargav5:
This is so nicely done @qiweiii @krijoh92 🙇🏽

PR #768, opened 2021-12-31
useIntervalWhen in useCountdown

trevorblades:
This resolves a deprecation notice in the console whenever I use `useCountdown`.

```
Warning: useInterval is deprecated, it will be removed in rooks v7. Please use useIntervalWhen instead. 
```


imbhargav5:
Thank you!! 🎉

imbhargav5:
Some tests seem to be failing though.

trevorblades:
It was failing due to me passing `null` to `useIntervalWhen` when it expected `number` or `undefined`. I just changed the `null` to `undefined`. Hopefully that works!

PR #785, opened 2022-01-13
Rerender session hook on remove

### LINKED ISSUE #780: useSessionstateStorage does not re-render when removing item from storage
**Describe the bug**
The `useSessionstateStorage` docs read:

> useState but syncs with sessionstorage

This implies that any change implemented with this hook will cause a re-render. At the moment, using the remove function does not cause the component to re-render.

If this is desired behavior, we should document it. 

If we want it to re-render (which makes sense to me), then I can look into making a PR.

LongLiveCHIEF:
fixes #780 

This removes the item from session storage, then internally calls the existing setState, setting it to `null`, casing the component to re-render, matching the behavior contract of `useState` in that the component re-renders each time the data is affected.

Note: This is a _breaking_ change.

### DIFF
--- a/src/hooks/useSessionstorageState.ts
+++ b/src/hooks/useSessionstorageState.ts
@@ -94,8 +94,8 @@ function useSessionstorageState<S>(
   }, []);
 
   const remove = useCallback(() => {
-    __setValue(null);
     sessionStorage.removeItem(key);
+    __setValue(null);
   }, []);
 
   return [value, setValue, remove];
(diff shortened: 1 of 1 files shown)


imbhargav5:
I might take some time to review this!

imbhargav5:
I would love it if we can somehow test this.

LongLiveCHIEF:
Yeah, that's been bugging me a bit too so I e been putting thought into it.

I might have an idea about how we could use the profiler api in a test to
achieve this.  Let me look at it later today and get back to you.

LongLiveCHIEF:
going to try and take a look at this today.

imbhargav5:
Closing this due to inactivity but please feel free to reopen if we have tests as this is a breaking change.

Issue #586, opened 2021-09-06
useStackState and useQueueState are using `any`. we should instead use generics here.


imbhargav5:

mrdulin:
Hi. I made a [PR](https://github.com/imbhargav5/rooks/pull/790)

PR #790, opened 2022-01-19
feat: add more methods for useStackState

mrdulin:
Changes to `useStackState` hook:

1. Added `clear` and `isEmpty` methods. 
2. Added unit testings for the above methods.
3. Use `list.length` instead of `length` in the closure to solve this eslint warning.
4. Use generics instead of `any` type. Related issue https://github.com/imbhargav5/rooks/issues/586


imbhargav5:
Incredible work! Love it @mrdulin !

PR #804, opened 2022-01-24
feat: add more unit tests for useFreshRef

mrdulin:
1. Added more unit tests for `useFreshRef`
2. Improve TS types.


src/hooks/useFreshRef.ts
```diff
@@ -14,7 +14,7 @@ import { useIsomorphicEffect } from './useIsomorphicEffect';
 function useFreshRef<T>(
   value: T,
   preferLayoutEffect: boolean = false
-): RefObject<T> {
+): MutableRefObject<T> {
```
imbhargav5, line 17: Ah. This is to avoid `current: T | null` and instead do `current: T`.

PR #823, opened 2022-02-19
Disable useWindowEventListener warning logs in production.

AbdallahAbis:
This hook currently clutters logs on SSR.

### DIFF
--- a/src/hooks/useWindowEventListener.ts
+++ b/src/hooks/useWindowEventListener.ts
@@ -27,9 +27,10 @@ function useWindowEventListener(
       isLayoutEffect
     );
   } else {
-    console.warn(
-      "useWindowEventListener can't attach an event listener as window is undefined."
-    );
+    if (process.env.NODE_ENV==='development')
+      console.warn(
+        "useWindowEventListener can't attach an event listener as window is undefined."
+      );
   }
 }
 


imbhargav5:
Ah. This is a great catch. We actually have a hook for this. I would appreciate greatly if you use the `useWarningOnMountInDevelopment ` and just put this message to log in development on mount.

AbdallahAbis:
@imbhargav5 I don't see any docs for the hook, and also Rooks don't have a named export `useWarningOnMountInDevelopment`?

imbhargav5:
Ah I should have explained this better. Apologies. We have an internal util . It's not for external use. Hence not exported. I am on mobile, let me link the actual file here.

imbhargav5:
Here is the link - https://github.com/imbhargav5/rooks/blob/c669a8eb841b0d74218365526b5182adae658443/src/hooks/useWarningOnMountInDevelopment.ts

imbhargav5:
On further thought, I think that warning itself is completely useless as it is running in the render phase. I think it's best to remove the warning and warn within `useGlobalEventListener` effect phase.

imbhargav5:
https://github.com/imbhargav5/rooks/issues/836 Here is the issue corresponding to this. If you would like to contribute, please take a shot, else, I will get to it tomorrow! Many thanks!

imbhargav5:
Closing as this is resolved.

PR #889, opened 2022-03-21
Able to stop observing with nullish callback

hooriza:
There is no way to stop intersection observing now.
So I want to able to it with this change.


imbhargav5:
This makes sense. But I am just going to think over the API for a bit and see if there is anything else we can do here or come up with a slightly different API.

imbhargav5:
Cool this makes sense to me.

PR #897, opened 2022-03-23
Change callback param type definition to optional

superLipbalm:
In the changed `useIntersectionObserverRef` [#889](https://github.com/imbhargav5/rooks/pull/889),
observation could be stopped by passing undefined to callback,
but callback was not optional, so undefined could not be passed in TS.
So I fixed the type definition.

### DIFF
--- a/src/hooks/useIntersectionObserverRef.ts
+++ b/src/hooks/useIntersectionObserverRef.ts
@@ -17,7 +17,7 @@ const config: IntersectionObserverInit = {
  * @param {IntersectionObserverInit} options
  */
 function useIntersectionObserverRef(
-  callback?: IntersectionObserverCallback,
+  callback: IntersectionObserverCallback | undefined,
   options: IntersectionObserverInit = config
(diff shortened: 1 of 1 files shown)


src/hooks/useIntersectionObserverRef.ts
```diff
@@ -17,7 +17,7 @@ const config: IntersectionObserverInit = {
  * @param {IntersectionObserverInit} options
  */
 function useIntersectionObserverRef(
-  callback: IntersectionObserverCallback,
+  callback?: IntersectionObserverCallback,
```
imbhargav5, line 20: Can we instead change this to `callback: IntersectionObserverCallback | undefined` ? This hook is practically useless without the callback but since we want to handle an `undefined` use case, let's put the types to match it?

superLipbalm: It looks good to be able to show more clearly what we want. I got it!

PR #974, opened 2022-05-02
Fix useDidUpdate to make it work with Strict Effects

ssmkhrj:
> With the release of React 18, StrictMode gets an additional behavior to ensure it's compatible with reusable state. When StrictMode is enabled, **React intentionally double-invokes effects (mount -> unmount -> mount) for newly mounted components**. [Source](https://github.com/reactwg/react-18/discussions/19)

So, it's essential that the `hasMountedRef` is reset when unmounting otherwise it won't behave desirably when wrapped inside StrictMode.

### DIFF
--- a/src/hooks/useDidUpdate.ts
+++ b/src/hooks/useDidUpdate.ts
@@ -26,6 +26,9 @@ function useDidUpdate(callback: () => any, conditions?: any[]): void {
     } else {
       hasMountedRef.current = true;
     }
+    return () => {
+      hasMountedRef.current = false;
+    }
   }, conditions);
 }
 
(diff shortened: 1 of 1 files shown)


ssmkhrj:
Closing this because I don't think this is the correct approach (even though it works) because it isn't guaranteed that all effects would run the same no. of time. I think that this hook doesn't even make sense now because this hook distinguishes between mount and updates but React doesn't distinguish between the two, and now it won't even let others distinguish between the two.

imbhargav5:
Wow. Thanks for this! Will review this tonight!

ssmkhrj:
@imbhargav5 Hey, did you manage to review this?

imbhargav5:
Hi @ssmkhrj I haven't had much of a chance to test the docs and be sure of approving this. What would really help me is if we added some tests to show the exact behaviour like some of our other hooks. Could I bother you to add a few tests for this ? That would help immensely.

imbhargav5 (approved):

Issue #877, opened 2022-03-08
[useFullscreen] Return type appears incorrect


bthall16:
https://github.com/imbhargav5/rooks/blob/49e14919944b7755926644cc0af7a0c05ff61f1d/src/hooks/useFullscreen.ts#L150

The return type for the `useFullscreen` hook is of type `FullscreenApi | undefined`. However, looking through the function body I don't see any path where `undefined` is returned. Consequently, TypeScript users have to perform type checks or add non-null assertions, both of which are unnecessary.

qqpann:
@imbhargav5 This seems to be already fixed.

imbhargav5:
@qqpann Got it. Will close 👍

PR #1006, opened 2022-05-19
fix: change useFullscreen hook return type (it never returns undefined)

sszczep:
As per #877, the useFullscreen hook has an incorrect typing of a return value.


sszczep:
Also i think that `element` arg in `toggle` and `request` methods should be typed as HTMLElement | undefined | null. As of now, we cannot directly pass React.RefObject<HTMLElement>.current as it resolves to HTMLElement | null.

imbhargav5:
Thanks for bringing this up. We are actually rewriting useFullscreen from scratch for the next release and I understand your frustration with it's typing.

Issue #1016, opened 2022-05-25
[TS] useForkRef type definition is incompatible with null or specific type of html element


neolivz:
**Describe the bug**

Type definition of `useForkRef` does not support null value to be passed as one of the ref. This is problematic as using with forwarded ref which can be null
```
    type ForwardedRef<T> = ((instance: T | null) => void) | MutableRefObject<T | null> | null;
```
There is no type safety check for making sure that the combined ref is used only on a particular type of HTMLElement, e.g HTMLButtonElement.

**To Reproduce**
Steps to reproduce the behavior
```
const Component = forwardRef<HTMLButtonElement, { callback: () => void }>(
  ({ callback }, ref) => {
    const [myIntersectionObserverRef] = useIntersectionObserverRef(callback)
    // Should only work for Button
    const newRef = useForkRef(ref, myIntersectionObserverRef)

    return <div ref={newRef}></div>
  }
)
```

**Expected behavior**
`useForkRef` should not throw error for ref being null
assignment of `ref={newRef}` should throw error because the element is not button but a div

**Screenshots**
![image](https://user-images.githubusercontent.com/754818/170154937-94faa358-b257-465b-bad1-3d245b579bf1.png)

qqpann:
@imbhargav5 Is this closed, since it is related to a PR?

PR #1017, opened 2022-05-25
fix: [1016] updated type definition of useForkRef with generic suppor…

neolivz:
…t and null support
Bug #1016

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,6 +1,5 @@
 {
   "trailingComma": "es5",
   "tabWidth": 2,
-  "singleQuote": true,
   "semi": true
 }
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "tabWidth": 2,
+  "singleQuote": true,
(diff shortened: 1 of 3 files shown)


.prettierrc.json
```diff
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
   "tabWidth": 2,
+  "singleQuote": true,
```
imbhargav5, line 4: @neolivz Can we remove this?

src/hooks/useForkRef.ts
```diff
@@ -19,7 +19,10 @@ function setRef(ref: AnyRef, value: HTMLElementOrNull) {
  * @param refA
  * @param refB
  */
-function useForkRef(refA: AnyRef, refB: AnyRef): CallbackRef | null {
+function useForkRef<T extends HTMLElement | null = HTMLElementOrNull>(
```
imbhargav5, line 22: Amazing!

.prettierrc.json
imbhargav5, line 4:
```suggestion
```

PR #1030, opened 2022-06-17
fix: Remove console.warn in SSR

### LINKED ISSUE #1029: Remove console.warn in SSR.
Hi.

I would like to request the removal of the `console.warn` in https://github.com/imbhargav5/rooks/blob/main/src/hooks/useOnWindowResize.ts#L28 and https://github.com/imbhargav5/rooks/blob/main/src/hooks/useOnWindowScroll.ts#L27

This log gets printed during Server Side Render. It's obvious that there is no window object in SSR and that this hook should do nothing. Having this printed in the terminal 10+ times every render is not needed.

qqpann:
The warning was unnecessary for several reasons.

First, The only way to detect SSR seemed to be using `window === undefined`, so detecting it and warning it is redundant.
Reference: https://github.com/alex-cory/use-ssr/blob/master/useSSR.ts

Secondly, in other place rooks calling `useGlobalObjectEventListener`, it does not detect `window === undefined`.
Reference:
https://github.com/imbhargav5/rooks/blob/bcd25dff395a82d51d27d4e1faac3bbbfe1dcd08/src/hooks/useDocumentEventListener.ts#L20
https://github.com/imbhargav5/rooks/blob/bcd25dff395a82d51d27d4e1faac3bbbfe1dcd08/src/hooks/useWindowEventListener.ts#L20

Finally, it violates conditional hooks call rule. The eslint shows error `React Hooks must be called in the exact same order in every component render.`
Reference: https://reactjs.org/docs/hooks-rules.html#gatsby-focus-wrapper

---
close #1029

### DIFF
--- a/src/__tests__/generalSsrTests.spec.ts
+++ b/src/__tests__/generalSsrTests.spec.ts
@@ -6,8 +6,6 @@ import { useFullscreen } from "../hooks/useFullscreen";
 import { useIntervalWhen } from "../hooks/useIntervalWhen";
 import { useLocalstorage } from "../hooks/useLocalstorage";
 import { useLocalstorageState } from "../hooks/useLocalstorageState";
-import { useOnWindowResize } from "../hooks/useOnWindowResize";
-import { useOnWindowScroll } from "../hooks/useOnWindowScroll";
 import { useOnline } from "../hooks/useOnline";
 import { useSessionstorage } from "../hooks/useSessionstorage";
 import { useThrottle } from "../hooks/useThrottle";
@@ -21,16 +19,6 @@ describe("when window is undefined", () => {
 
   afterEach(() => {
     consoleSpy.mockClear();
-  });
-
-  it("useOnWindowResize logs warning", () => {
-    renderHook(() => useOnWindowResize(mockCallback));
-    expect(consoleSpy).toHaveBeenCalledTimes(1);
-  });
-
-  it("useOnWindowScroll logs warning", () => {
-    renderHook(() => useOnWindowScroll(mockCallback));
-    expect(consoleSpy).toHaveBeenCalledTimes(1);
   });
 
   it("useIntervalWhen logs warning", () => {
--- a/src/hooks/useOnWindowResize.ts
+++ b/src/hooks/useOnWindowResize.ts
@@ -15,20 +15,14 @@ function useOnWindowResize(
   when: boolean = true,
   isLayoutEffect: boolean = false
 ) {
-  if (typeof window !== "undefined") {
-    useGlobalObjectEventListener(
-      window,
-      "resize",
-      callback,
-      { passive: true },
-      when,
-      isLayoutEffect
-    );
-  } else {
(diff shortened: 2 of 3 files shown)


qqpann:
@imbhargav5 Hi, I saw the CI failed and realized the test failure and fixed it.

imbhargav5:
@qqpann Thank you!

jekh:
This is actually causing errors in SSR. Since the `typeof window === "undefined"` guard has been removed, references to `window` cause SSR to throw a `ReferenceError` that completely breaks rendering instead of log an annoying-but-harmless error.

A good solution that both fixes SSR and follows the react hooks rule of "always call the same hooks in the same order on every render" would be to allow passing `undefined` anywhere `window` is required, and use a ternary like this:
```ts
  useGlobalObjectEventListener(
    window: typeof window !== "undefined" ? window : undefined,
    eventName,
    callback,
    listenerOptions,
    true,
    isLayoutEffect
  );
```

Then in [`useGlobalObjectEventListener`](https://github.com/imbhargav5/rooks/blob/bcd25dff395a82d51d27d4e1faac3bbbfe1dcd08/src/hooks/useGlobalObjectEventListener.ts#L19), simply do nothing if `globalObject` is undefined (which is common in SSR and not a problem). No need to log a warning there.

@imbhargav5  In the meantime, could this PR be reverted so SSR works again? This breaks a _lot_ of code.

imbhargav5:
👀

imbhargav5:
@jekh On it!

jekh:
Wow, thank you!

One thought on the warning in [`useGlobalObjectEventListener`](https://github.com/imbhargav5/rooks/blob/bcd25dff395a82d51d27d4e1faac3bbbfe1dcd08/src/hooks/useGlobalObjectEventListener.ts#L19): If you want to preserve the warning when code running in the browser accidentally passes `undefined` but avoid the unnecessary warnings for SSR, you could wrap the warning statement at [line 33](https://github.com/imbhargav5/rooks/blob/bcd25dff395a82d51d27d4e1faac3bbbfe1dcd08/src/hooks/useGlobalObjectEventListener.ts#L33) with a `typeof window` guard:

```ts
if (typeof window !== "undefined") {
  warning(
   typeof globalObject !== "undefined",
      "[useGlobalObjectEventListener]: Cannot attach event handlers to undefined."
  )
}
```

It looks like `useGlobalObjectEventListener` already properly handles its `globalObject` parameter being undefined, which is of course why this previously worked for SSR, just with the warning.

qqpann:
Sorry about the trouble. 😰Accepting undefined sounds to be a right direction

imbhargav5:
@qpann Don't worry about it 😉. I broke many many things. This is a super small thing. 
In fact do you want to try above the suggestion regarding warnings?

sgehrman:
does this work?  I'm having this issue

imbhargav5:
@sgehrman Going to fix this one tonight. Thanks for reporting everyone.

PR #1070, opened 2022-06-30
Better typing for useGlobalEventListener

### LINKED ISSUE #581: Better typing for useGlobalEventListener

qqpann:
I added typing to useGlobalEventListener and its related hooks.

---
closes #581


qqpann:
I reverted some eslint format, because Codeconv dropped.
The eslint config may require some improvements, but coverage-drop does not make sense to me 🤔

qqpann:
Looks like "type modifier on named imports" is available in typescript 4.5 or later, but the repository is using typescript 4.3. Even though the eslint is trying to use it, and causing conflict.

imbhargav5:
> I didn't update the test because lint fix will create too many changes. I would like to update them if it is ok with you (@imbhargav5 )
@qqpann So perhaps the lint fix can be a separate PR ? What do you think?

qqpann:
> So perhaps the lint fix can be a separate PR ? What do you think?

A separate PR sounds good to me!

imbhargav5:
@qqpann This looks like a great addition and I think shouldn't break anything. Do you want to add that in this PR?

qqpann:
Sure!

src/hooks/useWindowEventListener.ts
```diff
@@ -7,14 +7,21 @@ import { useGlobalObjectEventListener } from "./useGlobalObjectEventListener";
...
- * @param {object} conditions The options to be passed to the event listener
+ * @param {boolean | {once?: boolean; passive?: boolean; signal?: AbortSignal; capture?: boolean;}} listenerOptions The options to be passed to the event listener
  * @param {boolean} isLayoutEffect Should it use layout effect. Defaults to false
  * @returns {undefined}
  */
 function useWindowEventListener(
   eventName: string,
   callback: (...args: any) => void,
-  listenerOptions: any = {},
+  listenerOptions:
```
imbhargav5, line 17: Since this is less readable, what do you think of creating a type in `src/types/utils.ts` and then refer that here? 

So perhaps something like 

```
type EventListernerOptions =  boolean
    | {
        capture?: boolean;
        once?: boolean;
        passive?: boolean;
        signal?: AbortSignal;
      } = {},
```

And the import it here?

qqpann: I thought about that, thanks for pointing out!
Sounds great, will do it!

imbhargav5: I wonder if we can refer to the same type in jsdoc comment easily from src/types/utils 🤔 .

src/hooks/useGlobalObjectEventListener.ts
```diff
@@ -11,7 +11,7 @@ import { warning } from "./warning";
  * @param {Window|Document} globalObject The global object to add event onto
  * @param {string} eventName The event to track
  * @param {Function} callback The callback to be called on event
- * @param {object} conditions The options to be passed to the event listener
+ * @param {boolean | {once?: boolean; passive?: boolean; signal?: AbortSignal; capture?: boolean;}} listenerOptions The options to be passed to the event listener
```
imbhargav5, line 14: @qqpann Is there a way to make this also import-able from src/types/utils? 🤔 I am not sure and wondering if you have an idea.

src/hooks/useWindowEventListener.ts
qqpann: Seems `EventListenerOptions` is declared in typescript, so I'll name it `ListenerOptions`.

src/hooks/useGlobalObjectEventListener.ts
qqpann: Hmm, to my knowledge, we can write like this:

```ts
/**
  * @param {ListenerOptions} listenerOptions The options to be passed to the event listener
  */
```

but the actual type inside will be hidden by this name alias.

I think this is a matter of preference.

qqpann: But if we create an alias type already, it's cleaner to use it here as well IMO.

imbhargav5: Yeah this is fine. I want the type to be aliased, so if JSDOC type is hidden, that is fine.

src/hooks/useWindowEventListener.ts
qqpann: The GitHub did not update the chat and I didn't saw your comment.
Do you mean to put the type in src/types/utils?

qqpann: Sounds great!
I just didn't know the directory structure.

imbhargav5: image
This one. This is a new file that I added today. You might want to pull `main` and rebase.

qqpann: Got it!

qqpann:
@imbhargav5 I made the changes.
But I don't know why coverage drops this time 😅

imbhargav5:
@qqpann Don't worry about it. I think codecov is faulty and needs to be improved. I am happy with the PR. Thanks! 🙇🏽

imbhargav5 (approved):

PR #1140, opened 2022-07-11
fix(useDebounce)

Oszoou:
adds const callbackRef = useRef<T>(callback);

sandbox: https://codesandbox.io/s/usedebounce-forked-6sq8xh
see also: https://stackoverflow.com/questions/59151707/debounce-triggering-multiple-times-in-react

rooks versions <= 5.9.0 did not had that issue

### DIFF
--- a/src/hooks/useDebounce.ts
+++ b/src/hooks/useDebounce.ts
@@ -26,13 +26,18 @@ function useDebounce<T extends (...args: any[]) => unknown>(
     [wait, options]
   );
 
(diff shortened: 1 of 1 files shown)


imbhargav5:
@Oszoou thanks for this PR!  I definitely see the issue now. Could you help me further by adding some test cases? Would appreciate it very much!

imbhargav5:
@Oszoou I made a few changes as part of a larger refactor. Could you please confirm if this is still an issue? 🙏🏽

PR #1141, opened 2022-07-12
docs: Update useRaf to document timeElapsed

gpoole:
Updates the documentation for useRaf to specify what the timeElapsed value is and demonstrate how it might be used.


imbhargav5:
@gpoole Thanks!

Issue #1012, opened 2022-05-22
`yarn new` command error to create new hook


ynifamily3:
**Describe the bug**
I tried to contribute by creating a new hook according to the guide in the documentation, but I found the following type error:

```bash
/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:513
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
helpers/create/index.ts:65:53 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string | ReplacementFunction'.
  Type 'undefined' is not assignable to type 'string | ReplacementFunction'.

65   result = replaceString(result, "%directoryName%", directoryName);
                                                       ~~~~~~~~~~~~~
helpers/create/index.ts:66:51 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string | ReplacementFunction'.

66   result = replaceString(result, "%packageName%", packageName);
                                                     ~~~~~~~~~~~

    at createTSError (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:513:12)
    at reportTSError (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:517:19)
    at getOutput (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:752:36)
    at Object.compile (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:968:32)
    at Module.m._compile (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:1056:42)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Object.require.extensions.<computed> [as .ts] (/home/****/workspace/rooks/node_modules/ts-node/src/index.ts:1059:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
```

**To Reproduce**
1. Clone this repository
2. `yarn`
3. `yarn new`

**Expected behavior**
It should be ready to create a new hook without errors.

**Screenshots**
<img width="808" alt="image" src="https://user-images.githubusercontent.com/13795765/169691471-4df0c031-ce2d-4607-a114-79028304047c.png">

**Desktop (please complete the following information):**
[...]

PR #1143, opened 2022-07-13
Fix helpers/create

akiszka:
This pr fixes errors in helpers/create (`yarn new`) as described in #1012 

It:
- adds `?? ""` in two places to remove the possibility of a value being undefined
- removes the dependency on replace-string, adding a custom implementation

Questions for maintainers:
is helpers/create/index.ts the right location for a custom replaceString? It is also referenced in another helper (helpers/update-title-card/index.ts), so it might be more appropriate to place it in a separate file.

### DIFF
--- a/helpers/create/index.ts
+++ b/helpers/create/index.ts
@@ -2,13 +2,22 @@ import inquirer from "inquirer";
 import fs from "fs";
 import fsExtra from "fs-extra";
 import path from "path";
-import replaceString from "replace-string";
-import makeDir from "make-dir";
-import ora from "ora";
-import execa from "execa";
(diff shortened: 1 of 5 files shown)


imbhargav5:
Looks great!

imbhargav5:
Thanks @akiszka !

imbhargav5:
> is helpers/create/index.ts the right location for a custom replaceString? It is also referenced in another helper (helpers/update-title-card/index.ts), so it might be more appropriate to place it in a separate file.
@akiszka That's a great callout. It makes sense to move it into a separate file. However, `update-title-card`, is probably no longer needed. I will need to cross check that.

Issue #1142, opened 2022-07-12
useAsyncEffect


akiszka:
Hello,

I would propose to add a useAsyncEffect hook to rooks. This hook would be a simple clone of useEffect that accepts an async function.

I can write the hook and create a pr, but I would first like to know if there is demand for this kind of thing.

# Rationale

I often find myself writing lots of boilerplate code when running an async function inside useEffect. UseEffect only works with effect functions that have the type signature `() => (void | Destructor)`, so this is unacceptable:

`useEffect(async () => { ... })`

Because of this I find myself writing code similar to:

```jsx
useEffect(() => {
    (async () => {
        ...
    })()
})
``` 

which can get messy quickly, especially for short functions, or code like:

```jsx
useEffect(() => {
    doSomething()

    async function doSomething() {
        ...
    }
})
```

which looks relatively innocent, but note that the effect function now has 4 extra lines. This can quickly add up, especially when you call 3+ useEffects in a component.

All this could be replaced with a simple hook like:

```jsx
useAsyncEffect(async () => {
    ...
})
```

# Basic implementation

Below is an extremely basic implementation for context:

```tsx
import React, { useEffect } from "react";

type Effect = () => Promise<void>;
type CleanupFunction = () => void;

export function useAsyncEffect(effect: Effect, deps?: React.DependencyList, cleanup?: CleanupFunction) {
    useEffect(() => {
        effect()
        return cleanup
    }, deps)
}
```

imbhargav5:
@akiszka Looks like a great idea! Do you have any thoughts on how to clean up pending promises if a component un-mounts before the async task has finished? I guess we can use AbortController or something. What do you think?

imbhargav5:
@akiszka Yeah we should use AbortController. I am not sure we need cancellable option right now. We can make all of them cancellable on unmount and if there are specific repetitive edge cases we can take a call then.

qqpann:
This looks like good to close? @imbhargav5

imbhargav5:
Yep! I used @akiszka's code and then made some small changes and the hook is now live.

imbhargav5:
Thanks a lot @akiszka !

PR #1146, opened 2022-07-15
feat: Add useAsyncEffect

akiszka:
As per #1142, this PR adds the useAsyncEffect hook.

It turns out that cancelling promises with an AbortController is impossible in JS. Instead, this code gives the user the option to cancel their own effect function by providing an AbortSignal to them.

### DIFF
--- a/src/__tests__/useAsyncEffect.spec.ts
+++ b/src/__tests__/useAsyncEffect.spec.ts
@@ -101,6 +101,11 @@ describe("useAsyncEffect", () => {
 
     setTimeout(() => {
       expect(consoleSpy).toHaveBeenCalledTimes(1);
+      expect(consoleSpy).toHaveBeenCalledWith(
+        expect.stringContaining(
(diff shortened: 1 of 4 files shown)


src/__tests__/useAsyncEffect.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    renderHook(() => {
+      useAsyncEffect(async () => {
+        await new Promise((_, reject) => {
+          reject();
+        });
+      }, []);
+    });
+
+    setTimeout(() => {
+      expect(consoleSpy).toHaveBeenCalledTimes(1);
```
imbhargav5, line 103: Nitpick: It would be great if we can test some of the text from the warning here along with the number of times console was called. 🙇🏽

akiszka:
> I added test for the warning. BTW if you decide this is good to merge, you can also close #1142

imbhargav5:
Alright. Will get this shipped tonight!

PR #1151, opened 2022-07-23
fix: cleanup types

imbhargav5:


src/hooks/useLocalstorage.ts
```diff
@@ -73,12 +74,13 @@ function useLocalstorage(
   }, []);
 
   // eslint-disable-next-line consistent-return
-  const remove = useCallback(() => {
+  const remove = useCallback((): false | undefined => {
```
imbhargav5, line 77: This is just unfortunate typing. Need to fix this in the next major release.

src/hooks/useSessionstorage.ts
```diff
@@ -79,12 +80,13 @@ function useSessionstorage(
   }
 
   // eslint-disable-next-line consistent-return
-  function remove() {
+  function remove(): null | undefined {
```
imbhargav5, line 83: This is just unfortunate typing. Need to fix this in the next major release.

PR #1173, opened 2022-07-24
feat: Add useAsyncEffect (#1146)

imbhargav5:
* Write useAsyncEffect hook and test

* Add a new test for useAsyncEffect

* Docs

* Fix useAsyncEffect cleanup

* Add docs

* Test for error message


imbhargav5:
The more I see this hook, the more I think we should just be using suspense for this instead of effects and write very icky code.

PR #1210, opened 2022-08-04
fix: keyup cannot be detected when blocked by alert etc.

### LINKED ISSUE #1018: Usekeys - incorrectly triggered after multi key press.
**Describe the bug**
If you press multiple keys to trigger an event, that same event can be triggered again by pressing the last key. E.g. Click CtrlLeft + z to trigger an event, can be triggered again by only pressing z. 

**To Reproduce**
1. Go to the codesandbox on the rooks [useKeys](https://react-hooks.org/docs/useKeys) page [here](https://codesandbox.io/s/usekeys-uzlde?from-embed)
2. Press CtrlLeft + z to trigger the event.
3. Only press z the next time and see that the event is triggered again.

**Expected behavior**
The event should only be triggered if all the specified keys are pressed.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Edge
 - Version 101.0.1210.32

**Additional context**
I believe the problem arises from this code snippet in useKeys.ts
usekeys
When debugging, I found that only the last pressed key was being set to false in PressedKeyMapping, so if you pressed CtrlLeft + z, only z would get set back to false, while control would remain true. So only pressing z would trigger the event as control is still set to true.

qqpann:
useKeys seems to work fine except when interrupted by `alert`. 
After some investigation, it seems that `alert` will block js execution and there is no way but to overwrite `window.alert`.

Also, I could not find a way to include the alert in the test, so I could not add new test cases.

## References
- https://stackoverflow.com/a/4867114/8776028
- https://stackoverflow.com/a/13593319/8776028

---
close #1018

### DIFF
--- a/packages/rooks/src/hooks/useKeys.ts
+++ b/packages/rooks/src/hooks/useKeys.ts
@@ -134,6 +134,38 @@ function useKeys(
 
     return noop;
   }, [when, target, keysList, handleKeyDown, handleKeyUp]);
+
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      const originalAlert = window.alert;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      window.alert = (message?: any) => {
+        for (const identifier of keysList) {
+          PressedKeyMapping[identifier] = undefined;
+        }
+        return originalAlert(message);
+      };
+
+      const originalConfirm = window.confirm;
+      window.confirm = (message?: string | undefined) => {
+        for (const identifier of keysList) {
+          PressedKeyMapping[identifier] = undefined;
+        }
+        return originalConfirm(message);
+      };
+
+      const originalPrompt = window.prompt;
+      window.prompt = (
+        message?: string | undefined,
+        _default?: string | undefined
+      ) => {
+        for (const identifier of keysList) {
+          PressedKeyMapping[identifier] = undefined;
+        }
+        return originalPrompt(message, _default);
+      };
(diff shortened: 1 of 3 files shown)


packages/rooks/src/hooks/useKeys.ts
```diff
@@ -134,6 +134,38 @@ function useKeys(
 
     return noop;
   }, [when, target, keysList, handleKeyDown, handleKeyUp]);
+
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      const originalAlert = window.alert;
```
imbhargav5, line 140: @qqpann While this is hacky, I can see how this will get the job done, can we add a cleanup to this effect, that will revert window.alert back to original alert when this component unmounts?

imbhargav5: This is a pretty awesome solution, but I am just feeling yuck looking at this, because normally we would expect this to be taken care of by native browser, but we are having to come up with a solution like this to patch the browser. 😓 

Also if we can somehow make this opt-in, that would be even more awesome. Perhaps a param, that when set, will run this window.alert stuff.

qqpann: OK. Made these changes.

imbhargav5: @qqpann Man you are on 🔥 . Amazing!

imbhargav5 (approved):

PR #1214, opened 2022-08-04
chore: resolve eslint error

qqpann:
Resolved all eslint errors.

## Note

Could not solve useFullscreen document-related error, so I left `@ts-nocheck` as is.
I see the method is similar to `fscreen` (https://www.npmjs.com/package/fscreen), but I doubt is webkit- etc prefix is still needed.

### DIFF
--- a/packages/eslint-config-custom/index.js
+++ b/packages/eslint-config-custom/index.js
@@ -45,7 +45,6 @@ module.exports = {
     "import/no-unassigned-import": 0,
     "import/no-unresolved": 0,
     "max-nested-callbacks": 0,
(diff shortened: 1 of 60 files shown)


qqpann:
I checked by running `yarn --cwd packages/rooks lint`, but I'm not sure if the eslintrc is configured properly.

imbhargav5:
@qqpann HOLY!! What a PR!

imbhargav5:
@qqpann Solid PR. looks good. Yes we can definitely configure eslint better. Atleast we should find a way to integrate it with husky and lint-staged. This is awesome. Merging.

PR #1217, opened 2022-08-06
add: useKeys changeset

qqpann:
- added changeset for useKeys, which I forgot to include in https://github.com/imbhargav5/rooks/pull/1210
- updated the docs for useKeys

## Note

I can't find a workflow to use the updated rooks in the docs.
When I `yarn build` and `yarn dev` locally, it still seems to use the outdated rooks in the localhosted website.


imbhargav5:
@qqpann Fantastic!

PR #1226, opened 2022-08-08
feature: useFocus, useFocusWithin

### LINKED ISSUE #592: useFocus / useFocusWithin
An hook to detect if element has focus within and outside would be useful for accessibility, like the `:focus-within` css selector.

Something like this:
https://react-spectrum.adobe.com/react-aria/useFocusWithin.html

Using the react-aria hooks

qqpann:
Added two hooks: useFocus, useFocusWithin.

## References
- https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/test/useFocusWithin.test.js
- https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/test/useFocus.test.js
- https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/src/useFocusWithin.ts
- https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/src/useFocus.ts

## Note

I simplified them by removing conditions about "disabled" properties.

---
close #592

### DIFF
--- a/.changeset/fresh-readers-sit.md
+++ b/.changeset/fresh-readers-sit.md
@@ -1,5 +1,0 @@
----
-"rooks": patch
----
-
-useKeys comes with a new preventLostKeyup option for native events
--- a/.changeset/serious-glasses-check.md
+++ b/.changeset/serious-glasses-check.md
@@ -1,0 +1,5 @@
+---
+"rooks": minor
+---
+
+add useFocus & useFocusWithin hooks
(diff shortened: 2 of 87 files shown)


imbhargav5:
very clean! @qqpann

equinusocio:
imo the `disabled` property is useful when focus is added to non-focusable elements by using `tabindex`. In those cases there is no a native `disable` attribute that prevents keyboard focus by default like with `buttons` and other disabled elements.

PR #1231, opened 2022-08-09
fix: reduce package size and use path aliases

imbhargav5:


packages/rooks/.npmignore
```diff
@@ -0,0 +1,8 @@
+*
+!dist/**/*
+dist/esm/__tests__
```
imbhargav5, line 3: This file can probably be deleted.

.npmignore
```diff
@@ -1,7 +1,8 @@
 *
 !dist/**/*
-dist/esm/__tests__/**/*
-dist/cjs/__tests__/**/*
+dist/esm/__tests__
```
imbhargav5, line 3: This file can probably be deleted now.

imbhargav5:
@qqpann Thoughts on this PR?

.npmignore
qqpann: I see you have two `.npmignore` in the root and packages/rooks.
I guess we only need one of them

I checked with
```sh
# in packages/rooks
npm pack --dry-run
```
and removing either one of them seems to work fine.

qqpann: Is that what you meant with
> This file can probably be deleted now.

?

qqpann: @imbhargav5 What is your workflow to publish rooks to npm?
Do you use `npm publish` or let some tools handle it?

qqpann (approved):
Path aliases look pretty!

There is one remaining in the spec template FYI.

PR #1236, opened 2022-08-10
fix ssr errors in hooks that use window or document

gmahima:
**bug**: `window is not defined` (or `document is not defined`) error pops up when 
- `useDocumentEventListener`
- `useOnWindowScroll`
- `useOnWindowResize`
- `useWindowEventListener`
are called during SSR.

note: all these hooks call `useGlobalObjectEventListener`: 
```
useGlobalObjectEventListener(globalObject: Document | Window, eventName: string, callback: EventListener, listenerOptions?: ListenerOptions, when?: boolean, isLayoutEffect?: boolean): void
```

**fix**: add the following change to the hooks
call `useGlobalObjectEventListener` with `global.window` ( or`global.document`) instead of `window` (or `document`)

### DIFF
--- a/packages/rooks/src/hooks/useDocumentEventListener.ts
+++ b/packages/rooks/src/hooks/useDocumentEventListener.ts
@@ -18,20 +18,14 @@ function useDocumentEventListener(
   listenerOptions: ListenerOptions = {},
   isLayoutEffect = false
 ): void {
-  if (typeof document !== "undefined") {
-    /*
-    Since the above condition changes values only across different environments, it is fine to call the hook conditionally
-    */
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useGlobalObjectEventListener(
-      document,
-      eventName,
-      callback,
-      listenerOptions,
-      true,
-      isLayoutEffect
-    );
-  }
+  useGlobalObjectEventListener(
+    global.document,
+    eventName,
+    callback,
+    listenerOptions,
+    true,
+    isLayoutEffect
+  );
 }
 
 export { useDocumentEventListener };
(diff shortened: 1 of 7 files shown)


packages/rooks/src/hooks/useWindowScrollPosition.ts
```diff
@@ -48,6 +48,7 @@ function useWindowScrollPosition(): ScrollPosition {
    */
   useOnWindowResize(
     () => {
+      console.log("resize");
```
imbhargav5, line 51: This obviously shouldn't be here.

packages/rooks/src/hooks/useDocumentEventListener.ts
```diff
@@ -18,14 +18,20 @@ function useDocumentEventListener(
...
-    true,
-    isLayoutEffect
-  );
+  if (typeof document !== "undefined") {
+    /*
+    Since the above condition changes values only across different environments, it is fine to call the hook conditionally
+    */
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useGlobalObjectEventListener(
+      document,
```
imbhargav5, line 27: @gmahima  I just got this idea, instead of using a conditional, can you try just using `global.document` here?

gmahima: > @gmahima I just got this idea, instead of using a conditional, can you try just using `global.document` here?

It's working! It's working for the other hooks which use `window` too.

gmahima: But how is it working? Why is it not throwing an error? Isn't `global.window` and `window` the same object?

imbhargav5: It works because the reference it looks for is now `global` and if `global.window` doesn't exist, it will return undefined, instead of throwing an error.

gmahima: @imbhargav5 oh! ok

Issue #1235, opened 2022-08-10
[Suggestion] useFullscreen


sszczep:
Hey,

There have been a lot of issues with useFullscreen hook and I noticed that there is a rewritten version on one of the dev branches. I'm strongly against removing polyfill as stated in #579. The suggested polyfill in documentation ([fullscreen-polyfill](https://www.npmjs.com/package/fullscreen-polyfill)) won't work with Next.JS and other SSR. I'm also concerned about early return from hook (you even disabled the eslint rule). As I needed a solution to all above problems ASAP, I've created my own. 

[codesandbox](https://codesandbox.io/s/usefullscreen-8k389n?file=/src/useFullscreen.ts)
[demo](https://8k389n.csb.app) (fullscreen is not allowed in iframe on codesandbox)

I think it solves all the problems, is type-safe and also SSR friendly. 

Please let me know if you would be interested in including it in next major release or you'd rather stick with your implementation. Of course if it needs changes also let me know.

imbhargav5:
@sszczep Thanks for this! useFullscreen will be rewritten for sure for v7. It has a poor code quality and is buggy and the plan is to rewrite the whole thing in the next 2 weeks and I would greatly appreciate a PR!

Re: the polyfill. Is there a different polyfill that we can use that works with Next.js and other SSR libs? I really am not keen to increase bundle size by adding polyfills and if there is a way for users to choose their polyfill I want to leave it to them.

sszczep:
I don't think external polyfill is feasible, couldn't find any. 

I think the only solution is to include it using script tag (I really don't like that, what if there is a race condition and react mounts before the script executes? Seems error prone and requiring exhaustive testing) or by using next dynamic import (I don't like that either, it's platform specific and other renderers might not have such functionality. Also might still be impacted by race condition).

I stand by providing polyfill and not requiring users to manage it on their own. Our second option is to provide a Fullscreen API as a dependency. We would still provide our own polyfill, but allow users to use their own. Consider following code:

```ts
export interface FullscreenAPI {
  requestFullscreen: Element['requestFullscreen'];
  exitFullscreen: Document['exitFullscreen'];
  // ...
}

// Polyfilled Fullscreen API
const fullscreenAPI: FullscreenAPI = {
  requestFullscreen: (options?: FullscreenOptions | undefined) => Promise.resolve(),
  exitFullscreen: () => Promise.resolve(),
  // ...
};

// Returns hook with provided API
[...]

PR #1241, opened 2022-08-11
Rewrite useFullscreen hook

sszczep:
solves #1235

@imbhargav5 could someone takeover testing that hook? Removed test suite for now.


imbhargav5:
This looks great. Any thoughts @qqpann ? If it looks good to you, I will add tests and get it shipped this weekend.

packages/rooks/src/hooks/useFullscreen.ts
```diff
@@ -1,266 +1,226 @@
...
-    requestFullscreenOptions = {},
-  } = options;
-
-  const fullscreenControls = getFullscreenControls();
-  const [isFullscreen, setIsFullscreen] = useState(
-    Boolean(document[fullscreenControls.fullscreenElement])
-  );
-  const [element, setElement] = useState(
-    document[fullscreenControls.fullscreenElement]
+function useFullscreen(props: UseFullscreenProps = {}) {
```
qqpann, line 154:
```suggestion
/**
 *
 * useFullscreen hook
 *
 * Gives control to make HTML Elements fullscreen.
 *
 * @param {Element | undefined} props.target The target element to be fullscreen.
 * @param {(event: Event) => void} props.onChange The function to be called when the fullscreen changes.
 * @param {(event: Event) => void} props.onError The function to be called when the fullscreen error occurs.
 * @param {FullscreenOptions} props.requestFullscreenOptions The options to be passed to the requestFullscreen function.
 * @return {Object} returns - The controlls of useFullscreen hook.
 * @return {boolean} returns.isFullscreenAvailable - Whether the fullscreen is available.
 * @return {Element | null} returns.fullscreenElement - The fullscreen element.
 * @return {boolean} returns.isFullscreenEnabled - Whether the fullscreen is enabled.
 * @return {() => Promise<void>} returns.enableFullscreen - The function to enable fullscreen.
 * @return {() => Promise<void>} returns.disableFullscreen - The function to disable fullscreen.
 * @return {() => Promise<void>} returns.toggleFullscreen - The function to toggle fullscreen.
 */
function useFullscreen(props: UseFullscreenProps = {}) {
```

I tried adding JSDoc.

packages/rooks/src/hooks/useFullscreen.ts
```diff
@@ -1,266 +1,226 @@
...
-      for (const [index, _function] of functionSet.entries()) {
-        returnValue[functionMap[0][index]] = functionSet[index];
-      }
-    }
+import { useState, useCallback, useEffect } from "react";
+import type { RefObject } from "react";
+
+declare global {
+  interface Element {
+    webkitRequestFullscreen: Element["requestFullscreen"];
```
qqpann, line 6:
```ts
    webkitRequestFullscreen?: Element["requestFullscreen"];
```

Not sure if necessary, but how about making all of them undefined-able just to make it clear it is possibly undefined depending on the browser?

qqpann (approved):
I read through the code carefully and it looks amazing! 🔥

packages/rooks/src/hooks/useFullscreen.ts
sszczep: I would also add 

@param {Object} props Optional hook configuration

or something like that.

packages/rooks/src/hooks/useFullscreen.ts
sszczep: I was thinking about it. I couldnt find a way to replace existing type definitions (I wanted to make non-prefixed methods also optional). I didn't want to mix them. Seems more uniform this way. If there is a way to override it I can make further adjustments. Let me know what you think.

sszczep: I was also thinking about creating a separate type and casting document to it. Might be a better solution but I'm not sure if you like it.

sszczep: I'm also not sure how we would extend the event map that way.

packages/rooks/src/hooks/useFullscreen.ts
qqpann: I tried it at the beginning, but it will prevent the VSCode hovering window from showing what's inside of the props.

With `@param {Object}`:
スクリーンショット 2022-08-14 1 51 03

Without it:
スクリーンショット 2022-08-14 1 50 28

sszczep: I see. Looks good to me

packages/rooks/src/hooks/useFullscreen.ts
qqpann: I would generally prefer creating a separate type, but I have no idea what will that look like...

sszczep: I'll try to provide some examples in following days. Can @imbhargav5 start working on tests? Types should not interfere.

sszczep: Resolved in latest commit. I think it didn't turn out that bad.

sszczep:
One more thing - do you have any idea for better event name mapping? To be honest I don't like it much.

imbhargav5:
@sszczep @qqpann Awesome work! I will take a look tomorrow.

qqpann:
@imbhargav5 @sszczep Oh forgot to mention, it would be great if the documentation (useFullscreen.md) could be updated with the new API.

sszczep:
> @imbhargav5 @sszczep Oh forgot to mention, it would be great if the documentation (useFullscreen.md) could be updated with the new API.

I'll work on docs. Please think about that event name mapping. We also need unit tests

sszczep:
@imbhargav5 could you reopen? I can't push more commits. 

I wrote new documentation, however, I couldn't check how it renders. Should be fine though.

PR #1244, opened 2022-08-17
fix(useAsyncEffect): fix dependencies check

imbhargav5:


packages/rooks/src/hooks/useAsyncEffect.ts
```diff
@@ -48,8 +47,9 @@ function useAsyncEffect<T>(
     } catch (error) {
       throw error;
     }
+
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [effect, getIsMounted, deps]);
```
imbhargav5, line 52: This was causing continuous the callback to update too many times for no reason if the dependency was passed as `useAsynEffect(callback, [])` because the array is recreated on each render.

imbhargav5:
cc @qqpann

packages/rooks/src/hooks/useAsyncEffect.ts
qqpann: Interesting. Nice fix!

Issue #1362, opened 2022-09-01
[V7] Bundled umd does not pass es2015-es2019 checks


belgattitude:
The bundled `dist/umd/rooks.umd.js` contains an [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) which is probably too modern for some browsers. 

- see support https://caniuse.com/mdn-javascript_operators_optional_chaining

Not sure if it's intended as it does not appear in esm and cjs bundles. If it is would be interesting to mention it in the changelog and align the bundles.

![image](https://user-images.githubusercontent.com/259798/187943420-28480a47-1dc3-45c6-a137-15200b33ffba.png)

## Info

- PR: https://github.com/belgattitude/nextjs-monorepo-example/pull/2551
- Failing test: https://github.com/belgattitude/nextjs-monorepo-example/runs/8135920555?check_suite_focus=true

## To reproduce

clone the `renovate/rooks-7.x` branch from https://github.com/belgattitude/nextjs-monorepo-example

```bash
yarn install
cd apps/nextjs-app
yarn build && yarn check-dist
// or
yarn es-check es2016 "../../node_modules/rooks/dist/umd/**.js"
```

imbhargav5:
@belgattitude Thanks for this! Do you happen to know a solution to fix this easily? If you can submit a PR or guide us to a solution we can fix this quickly. Appreciate it! 🚀

belgattitude:
Sure I'll have a look this week-end.

belgattitude:
Hey @imbhargav5 now that I opened the repo, I see multiple things that would make sense to fix.

First I'd like to rework the exports field (would love to remove unpkg field too - that's the reason why nextjs picked up the umd build, rather than cjs or esm)

Second I feel for bundling esm/cjs I'd like to go with tsup if it makes sense to you. 

Third I'm wondering if I could remove the bundled test files.

I'll plan to make some PR's whenever I have time so you can follow and give me directions if needed.

Enjoy your day

imbhargav5:
@belgattitude Wow. Please go ahead and make a PR. I am not entirely sure of all the changes but it seems like you have a solid idea of the ideal setup. Once you make a PR based on your idea, I will do a thorough review and get it shipped asap.

PR #1365, opened 2022-09-02
chore: add check-size script

belgattitude:
**status** DO NOT MERGE, WILL be part of #1370

Used as reference to track possible size regressions in #1370 

image

PR #1370, opened 2022-09-03
Rework bundler and bring back support to es2017 

belgattitude:
## Purpose

Rework the bundler config to ensure emitted file are valid es2017 files. Will fix https://github.com/imbhargav5/rooks/issues/1362

## Approach

Tried to keep on rollup, but easily extensible for esbuild

## How to try it out

```basjh
cd ./packages/rooks
yarn build
yarn check:dist
yarn check:size
```

## Checks

Impact on size: negligible imo

image

## Tasks

- [x] Emit es2017 thx typescript (could work with esbuild too)
- [x] Add module export that might help some bundlers (no changes)
- [x] Removed unused files (ie: jest.setup.ts, test files)
- [x] Removed declaration maps  for dist (they are only useful in monorepo ?)
- [x] Removed dual types publishing (kept ./types/index.d.ts, removed per-file declarations)
- [x] Didn't understand what was the purpose of index.standalone, need to clear this
- [x] Merge rollup configs, and allow to build umd as well   - **To test it out**
- [x] Post-cleanup: remove obsoleted deps...


packages/rooks/package.json
```diff
@@ -7,6 +7,15 @@
   "unpkg": "dist/umd/rooks.umd.js",
   "types": "dist/types/index.d.ts",
   "typings": "dist/types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/types/index.d.js",
+      "module": "./dist/esm/index.js",
```
belgattitude, line 13: Kept the `module` for webpack support (it's not part of the final spec). Not sure if recent versions needs that.

Didn't add the `browser` field (umd), cause nextjs 12.x will pick it up and defeat tree-shakability (browser bundle is one file)

packages/rooks/rollup.config.ts
```diff
@@ -14,6 +14,7 @@ const commonPlugins = [
   sourceMaps(),
   json(),
   babel({
+    babelHelpers: "bundled",
```
belgattitude, line 17: removed a babel warning

packages/rooks/rollup.future.config.ts
```diff
@@ -0,0 +1,68 @@
...
+    plugins: [
+      tsPathPlugin,
+      dts({
+        compilerOptions: {
+          sourceMap: false,
+        },
+      }),
+    ],
+  },
+  // @todo UMD from rollup.config.ts
```
belgattitude, line 67: I'll merge from the existing rollup config

packages/rooks/rollup.future.config.ts
```diff
@@ -0,0 +1,68 @@
...
+const external = [
+  ...Object.keys(pkg.dependencies),
+  ...Object.keys(pkg.peerDependencies || {}),
+];
+
+const distDir = "./dist/future";
+
+const tsPathPlugin = typescriptPaths({
+  preserveExtensions: true,
+  tsConfigPath: "./tsconfig.json",
```
belgattitude, line 17: typescriptPaths plugin does not support extend from tsconfig.build.json. So I use the paths from there. Note that it makes use remove composite: true.

belgattitude:
Hey @imbhargav5. It's just a start.

But I'd like to know

- What's the purpose of publishing index.standalone.js ? Is it important ?
- Are the declaration sourcempas and sourcemaps important ?

I'm not totally sure about this.

Happy week-end.

packages/rooks/tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "module": "esnext",
     "jsx": "react",
     "target": "esnext",
-    "composite": true,
```
belgattitude, line 12: I had to remove this cause I will make the rollup-plugin-typescript-paths fail. See above comment.

I feel there's no need to set it, cause I haven't seen project references support

imbhargav5:
Going to look at this tonight!

packages/rooks/package.json
imbhargav5: @belgattitude Could you help me with any reading material for me to fully understand what has changed in the `exports` field? It changes so often I lost track of it. Just want to make sure I understand all the changes we are making and we don't accidentally break anything.

packages/rooks/tsconfig.json
imbhargav5: Got it. If we don't need it, let's scrap. 👍🏽

imbhargav5:
@belgattitude `index.standalone` is an artefact from a year ago and we don't need that anymore. Thanks for pointing that out. We can remove it. 😄

packages/rooks/package.json
belgattitude: Most up-to-date I found out is

https://github.com/frehner/modern-guide-to-packaging-js-library#define-your-exports

belgattitude:
@imbhargav5 

If all looks good to you, my plan is to replace the build with build-future and the new rollup config. I'll undraft when done.

My last concern, I saw a drop in size in the umd bundle. I guess is that the *.spec.ts aren't treated anymore. But I'd like another look.

packages/rooks/package.json
belgattitude: If you feel unsure I can cherry-pick this and create another PR.

imbhargav5:
Will take a look.

packages/rooks/package.json
imbhargav5: @belgattitude Thanks for this! I will read this and get back. I will take a good look at this PR tomorrow.

imbhargav5: TIL learnt a lot from that link today. Thanks!

imbhargav5:
@belgattitude  Looks like a great start. I am getting errors after I run `yarn run build:future && yarn run check:dist`. Is that expected?

belgattitude:
Yes it will check the ./dist/* folder, not the ./dist/future :smile: .  I'll P/R tomorrow.

belgattitude:
Hey @imbhargav5 I've just finalized the thing.  

Added 
- a (patch) changeset
- a probably useful 'typecheck' step (no on CI though), that can pave the way for using esbuild.

Entered pre canary mode, so If you don't mind  I'd like to test a canary version release first.

belgattitude:
Found a little time and fixed conflicts.

Ported the check-size in #1365 

Good news, the size increase is very minimal

image

The build passed from 20s to 15s on my laptop. Imo it can be merged, let me know if you have questions.

Thanks for this beautiful lib

packages/rooks/.size-limit.js
```diff
@@ -0,0 +1,42 @@
+// @ts-check
+
+// Regularly update this max size when new hooks are added (keep a threshold)
+const fullBundleMaxSize = "12KB";
```
belgattitude, line 4: @imbhargav5 just to mention, if new hooks are added and the total size of lib grow, this should be increased.

On the other side, it's totally possible to keep only the `"ESM Webpack ({ useDebounce })"` (see below), that's actually the important check

packages/rooks/package.json
```diff
@@ -7,6 +7,15 @@
   "unpkg": "dist/umd/rooks.umd.js",
   "types": "dist/types/index.d.ts",
   "typings": "dist/types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/types/index.d.js",
```
imbhargav5, line 12:
```suggestion
      "types": "./dist/types/index.d.ts",
```

packages/rooks/.size-limit.js
imbhargav5: @belgattitude Ah. I think we could write a small script to autopopulate new hooks into the size-limit file when new hooks are added and keep the limit at 2kb. What do you think?

imbhargav5:
This is such a massive improvement @belgattitude ! I just want to keep this in `pre` mode for a bit for a couple of days. Can we re-enter pre mode again?

imbhargav5:
@belgattitude I continued your work here. https://github.com/imbhargav5/rooks/pull/1408

packages/rooks/.size-limit.js
belgattitude: Nice idea

PR #1393, opened 2022-09-08
[useTimeTravelState] Fix: remove default export

belgattitude:
@imbhargav5 

Potentially a BC, so I'm not sure.

Detected when working on #1370. 

With default+named exports the esm/cjs bundles are a little more difficult to keep in line. There's a warning in the new bundler, so here's what I can do. 

- Either remove the default export (like in this P/R)
- Or add some config in rollup in #1370.

Let me know


.changeset/breezy-lemons-relate.md
```diff
@@ -0,0 +1,16 @@
...
+Remove useTimeTravelState default export
+
+The useTimeTravelState was exported as default and named exports which
+creates possible differences in our dual packaging (cjs, esm). 
+
+This is a potentially breaking change. If you're affected, please raise an issue or
+adapt the import as below.
+
+```typescript
+// Change 'import useTimeTravelState from "rooks"' into
```
imbhargav5, line 14: 🤔 I don't think this was the old import. It was probably something like 
`'import useTimeTravelState from "rooks/dist/esm/useTimeTravelState"'` which I don't think anyone used. So this should be fine to remove.

imbhargav5 (approved):

PR #1394, opened 2022-09-08
Changeset improvements (suggestion)

belgattitude:
I'd like to propose to use changeset to generate changelogs and release descriptions. Drop current scripts (yarn version).

The goal is to have more useful desc in the changelogs

-  PR link
- author
- ...

Will give more insights... 

@imbhargav5 do you have any thoughts about this ? I'm quite experienced with what I can do with changesets and open to exchange pro/cons, but don't really know the current system. Is there interest ?

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,6 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.1.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "changelog": [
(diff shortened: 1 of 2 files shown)


imbhargav5:
@belgattitude I have been meaning to do this for a long time. Thanks for taking this up! I like this approach and let's go for it. Just a headsup, I might be slightly slow to review your other PR this time. Will get it shipped this weekend for sure.

PR #1414, opened 2022-09-12
chore: add useasynceffect example

### LINKED ISSUE #1321: No example for useAsyncEffect
Currently, no example is found for [useAsyncEffect()](https://github.com/imbhargav5/rooks/blob/main/data/docs/useAsyncEffect.md?plain=1#L16)

Check out [the docs](https://react-hooks.org/docs/useAsyncEffect)

ddsuhaimi:
fixes #1321

### DIFF
--- a/data/docs/useAsyncEffect.md
+++ b/data/docs/useAsyncEffect.md
@@ -12,6 +12,23 @@ This is a version of `useEffect` that accepts an async function.
 
 [//]: # "Main"
 
-## Examples
+```jsx
+import { useDidMount } from "rooks";
+export default function App() {
+  const [value, setValue] = useState(false);
+  useAsyncEffect(async () => {
+    setTimeout(() => {
+      setValue(true);
+    }, 100);
+  }, []);
 
-### Basic example
+  return value;
+}
+```
+
+### Arguments
+
+| Argument | Type     | Description                 |
+| -------- | -------- | --------------------------- |
+| callback | function | Async function to be called |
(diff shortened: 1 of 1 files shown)


data/docs/useAsyncEffect.md
```diff
@@ -12,6 +12,23 @@ This is a version of `useEffect` that accepts an async function.
 
 [//]: # "Main"
 
-## Examples
+```jsx
+import { useDidMount } from "rooks";
```
imbhargav5, line 16: Let's import `useAsyncEffect` as well here.

data/docs/useAsyncEffect.md
```diff
@@ -12,6 +12,23 @@ This is a version of `useEffect` that accepts an async function.
...
+export default function App() {
+  const [value, setValue] = useState(false);
+  useAsyncEffect(async () => {
+    setTimeout(() => {
```
imbhargav5, line 20: Since this is an `async` effect, we want to use a code snippet which can `await` .

ddsuhaimi:
```jsx
import { useAsyncEffect  } from "rooks";

const foo = new Promise((resolve, reject) => {
  setTimeout(() => {
    resolve("bar");
  }, 300);
});

export default function App() {
  const [value, setValue] = useState(false);
  useAsyncEffect(async () => {
    await foo() 
  }, []);

  return value;
}
```

something like this?

PR #1497, opened 2022-10-04
Update useSelectableList.ts

### LINKED ISSUE #1496: useSelectableList does not support object list
There is a bug in useSelectableList, currently only number/string are supported as object list.

https://github.com/imbhargav5/rooks/blob/main/packages/rooks/src/hooks/useSelectableList.ts#L60

it could be fixed with `find` but its unknown which identifier is the `value`.

dohomi:
Potentially fix an array of objects closes #1496

### DIFF
--- a/packages/rooks/src/hooks/useSelectableList.ts
+++ b/packages/rooks/src/hooks/useSelectableList.ts
@@ -57,7 +57,7 @@ function useSelectableList<T>(
         if (typeof index !== "undefined") {
           setCurrentIndex(index);
         } else if (typeof value !== "undefined") {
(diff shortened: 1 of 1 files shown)


imbhargav5:
@dohomi This is closed though. How should we proceed on this issue in your opinion?

dohomi:
@imbhargav5 I closed it because I realised that actually the options needs to be bind with a `useMemo` and then everything seemed to work

imbhargav5:
@dohomi  I see. Let me know if there is any action needed on my end to make it easy for everyone. Will a useMemo within the hook make things easy?

PR #1560, opened 2022-10-27
use Element type instead of HTMLElement

### LINKED ISSUE #1557: Expand HTMLElementOrNull type to include SVGElement
I'm using useBoundingclientrectRef with an `<svg>` element, but I have to cast the ref to `RefCallback<SVGSVGElement>` in order to shut TypeScript up.  It would be great if you could expand that type to include SVGElement as well.
I'm guessing pretty much everything you're doing with HTMLElement is something that also works with SVGElement.

Thanks for a great lib!

jrnail23:
resolves #1557 
_Hopefully_ these changes are non-breaking.
Since both `HTMLElement` and `SVGElement` extend `Element`, and (at least according to TypeScript) all previous `HTMLElement` references are supported by `Element`.

Although I guess I'm not quite sure what users are doing with those refs, etc.

### DIFF
--- a/packages/rooks/src/hooks/useBoundingclientrect.ts
+++ b/packages/rooks/src/hooks/useBoundingclientrect.ts
@@ -1,3 +1,4 @@
+import { ElementOrNull } from "@/utils/utils";
 import type { MutableRefObject } from "react";
 import { useState, useCallback } from "react";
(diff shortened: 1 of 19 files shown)


imbhargav5:
@jrnail23 Thanks for this PR! I will take a look at this tomorrow.

PR #1564, opened 2022-11-28
useMediaMatch handles case where window is not defined

patik:
In some contexts (e.g. while running the Next.js dev server), `window` will not be defined, which causes this hook to throw an exception. This pull request protects against this situation; the hook returns `false` in that case (as in, the media query is not a match).

[Here's a Codesandbox](https://codesandbox.io/p/sandbox/cold-shadow-dpcsv6?file=%2Fpages%2Findex.tsx&selection=%5B%7B%22endColumn%22%3A55%2C%22endLineNumber%22%3A6%2C%22startColumn%22%3A55%2C%22startLineNumber%22%3A6%7D%5D) with the current (unpatched) hook running within Next.

### DIFF
--- a/packages/rooks/src/hooks/useMediaMatch.ts
+++ b/packages/rooks/src/hooks/useMediaMatch.ts
@@ -11,7 +11,7 @@ import { useEffect, useMemo, useState } from "react";
  * @see https://rooks.vercel.app/docs/useMediaMatch
  */
 function useMediaMatch(query: string): boolean {
-  const isWindowDefined = typeof window !== "undefined";
+  const isWindowDefined = typeof window !== "undefined" && window.matchMedia;
   const matchMedia = useMemo<MediaQueryList | undefined>(
     () => (isWindowDefined ? window.matchMedia(query) : undefined),
     [query, isWindowDefined]
@@ -21,7 +21,7 @@ function useMediaMatch(query: string): boolean {
   );
 
   useEffect(() => {
-    if (!matchMedia) {
+    if (!isWindowDefined || !matchMedia) {
       return;
     }
 
@@ -36,9 +36,9 @@ function useMediaMatch(query: string): boolean {
       matchMedia.addListener(listener);
       return () => matchMedia.removeListener(listener);
     }
-  }, [matchMedia]);
+  }, [isWindowDefined, matchMedia]);
 
-  if (typeof window === "undefined") {
+  if (isWindowDefined) {
     console.warn("useMediaMatch cannot function as window is undefined.");
 
     return false;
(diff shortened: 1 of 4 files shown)


patik:
I just added another commit because I found that during `jest --watch`, `window` can be reported as an `"object"` but it does not contain the `matchMedia` method. I would add tests for this condition but I'm not sure how.

packages/rooks/src/hooks/useMediaMatch.ts
```diff
@@ -11,13 +11,20 @@ import { useEffect, useMemo, useState } from "react";
...
+    () => (isWindowDefined ? window.matchMedia(query) : undefined),
+    [query, isWindowDefined]
+  );
+  const [matches, setMatches] = useState<boolean>(() =>
+    (isWindowDefined && matchMedia ? matchMedia.matches : false)
   );
-  const [matches, setMatches] = useState<boolean>(() => matchMedia.matches);
 
   useEffect(() => {
+    if (!isWindowDefined || !matchMedia) {
```
imbhargav5, line 24: I wonder if this is needed, since effects only run in the dom. Unless I am missing something.

nghiepdev:
I also have this issue

packages/rooks/src/hooks/useMediaMatch.ts
patik: When I remove that, the test no longer passes, and I'm not sure how to fix it. Here's the output:

```
rooks:test: Summary of all failing tests
rooks:test: FAIL __tests__/useMediaMatch.spec.tsx
rooks:test:   ● useMediaMatch › should track a boolean
rooks:test:
rooks:test:     expect(received).toBe(expected) // Object.is equality
rooks:test:
rooks:test:     Expected: true
rooks:test:     Received: false
rooks:test:
rooks:test:       51 |
rooks:test:       52 |     // Invoking the listener changes the value
rooks:test:     > 53 |     const l =
rooks:test:          |              ^
rooks:test:       54 |       expectDefined<(event_: MediaQueryListEventMap["change"]) => void>(
rooks:test:       55 |         listener
rooks:test:       56 |       );
rooks:test:
rooks:test:       at Object.<anonymous> (__tests__/useMediaMatch.spec.tsx:53:28)
```

And the relevant code:

```
  expect(result.current).toBe(true);

  // Invoking the listener changes the value
  const l =
    expectDefined<(event_: MediaQueryListEventMap["change"]) => void>(
      listener
    );
  act(() => l({ matches: false } as any));
  expect(result.current).toBe(false);
```

When it runs `l()` near the end of this snippet, since `window` is not defined, it returns the previous value of `isMatches` which was `true`.

PR #1567, opened 2022-12-31
add type support for useGlobalObjectEventListener

mass2527:
changed type of `useGlobalObjectEventListener` hook so that event type could be precisely inferred depending on the `eventName`


imbhargav5:
This looks good @mass2527! Just add this as a patch changeset and this will be good to go!

PR #1568, opened 2023-01-05
Reduce the frequency of `useIntersectionObserver` callback execution

hooriza:
Fixed an issue where the `IntersectionObserver` object was recreated and the callback was called excessively whenever the `options` or `callback` argument changed, even if the intersection state did not change.

#### AS-IS
```ts
const doSomethingWithEntry = useCallback(([entry]) => {
    // do something with entry
}, [ ...deps ]);

const callbackRef = useRef(doSomethingWithEntry);
callbackRef.current = doSomethingWithEntry;

const callback = useCallback((...args) => {
    return callbackRef.current?.(...args);
}, []);

const ref = useIntersectionObserverRef(
    callback,
    useMemo({ threshold: 0.5 }, [])
);
```

#### TO-BE
````ts
const callback = useCallback(([entry]) => {
    // do something with entry
}, [ ...deps ]);

const ref = useIntersectionObserverRef(callback, { threshold: 0.5 });
```

### DIFF
--- a/packages/rooks/src/hooks/useIntersectionObserverRef.ts
+++ b/packages/rooks/src/hooks/useIntersectionObserverRef.ts
@@ -40,7 +40,7 @@ function useIntersectionObserverRef(
 
   useEffect(() => {
     // Create an observer instance linked to the callback function
-    if (node && callbackRef.current) {
+    if (node) {
       const observer = new IntersectionObserver(handleIntersectionObserver, {
         root,
         rootMargin,
--- a/packages/rooks/src/hooks/useIntersectionObserverRef.ts
+++ b/packages/rooks/src/hooks/useIntersectionObserverRef.ts
@@ -27,7 +27,9 @@ function useIntersectionObserverRef(
   const [node, setNode] = useState<HTMLElementOrNull>(null);
 
   const callbackRef = useRef(callback);
-  callbackRef.current = callback;
+  useEffect(() => {
+    callbackRef.current = callback;
+  });
 
   const handleIntersectionObserver = useCallback<IntersectionObserverCallback>(
     (...args) => {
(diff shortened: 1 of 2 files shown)


imbhargav5:
Ah. Nice catch!

packages/rooks/src/hooks/useIntersectionObserverRef.ts
```diff
@@ -26,10 +26,24 @@ function useIntersectionObserverRef(
 
   const [node, setNode] = useState<HTMLElementOrNull>(null);
 
+  const callbackRef = useRef(callback);
+  callbackRef.current = callback;
```
imbhargav5, line 30: This has to be updated in a separate useEffect

```suggestion
useEffect(() => {
    callbackRef.current = callback;
})
```

packages/rooks/src/hooks/useIntersectionObserverRef.ts
```diff
@@ -26,10 +26,24 @@ function useIntersectionObserverRef(
...
+      return callbackRef.current?.(...args);
+    },
+    []
+  );
+
   useEffect(() => {
     // Create an observer instance linked to the callback function
-    if (node && callback) {
-      const observer = new IntersectionObserver(callback, options);
+    if (node && callbackRef.current) {
```
imbhargav5, line 41: I think we shouldn't be checking for callbackRef.current here. 
```suggestion
    if (node) {
```

imbhargav5: The `handleIntersectionObserver` will take care of calling the function if it exists :)

imbhargav5 (changes requested):
Nice work! Just minor changes needed as mentioned in review. Let me know.

packages/rooks/src/hooks/useIntersectionObserverRef.ts
hooriza: Thanks for your review!

packages/rooks/src/hooks/useIntersectionObserverRef.ts
hooriza: Thanks! I changed it

PR #1569, opened 2023-01-05
Prevent excessive call of resize observer callback

superLipbalm:
The same problem with #1568 also existed with `useMutationObserver` and we fixed it.
This issue can also be seen in examples in the [official documentation](https://rooks.vercel.app/docs/useResizeObserverRef).

useResizeObserverRef

Since a new callback is created and passed every render, then the callback is called because new `ResizeObserver` is created, and since the callback is changing the state of `resizeCount`, so it re-renders infinitely.

Additionally, it would be nice to be able to cancel observing using a nullish callback like `useIntersectionObserver`, so I modified it the same way.(ref: #897)

### DIFF
--- a/packages/rooks/src/hooks/useResizeObserverRef.ts
+++ b/packages/rooks/src/hooks/useResizeObserverRef.ts
@@ -1,7 +1,6 @@
 import { noop } from "@/utils/noop";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import type { CallbackRef, HTMLElementOrNull } from "../utils/utils";
-import { useFreshTick } from "./useFreshTick";
 
 const config: ResizeObserverOptions = {
   box: "content-box",
@@ -14,7 +13,7 @@ const config: ResizeObserverOptions = {
  * Returns a resize observer for a React Ref and fires a callback
  * https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
  *
- * @param {ResizeObserverCallback} callback Function that needs to be fired on resize
+ * @param {ResizeObserverCallback} callback Function that needs to be fired on resize: ;
  * @param {ResizeObserverOptions} options An options object allowing you to set options for the observation
  * @returns {[CallbackRef]} callbackref
  * @see https://rooks.vercel.app/docs/useResizeObserverRef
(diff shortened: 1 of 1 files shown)


packages/rooks/src/hooks/useResizeObserverRef.ts
```diff
@@ -20,26 +19,35 @@ const config: ResizeObserverOptions = {
...
+  const handleResizeObserver = useCallback<ResizeObserverCallback>(
+    (...args) => {
+      callbackRef.current?.(...args);
+    },
+    []
+  );
 
   useEffect(() => {
-    if (node) {
+    if (node && callbackRef.current) {
```
imbhargav5, line 38: Looks great. I don't think we need to check for `callbackRef.current` here

```suggestion
    if (node) {
```

packages/rooks/src/hooks/useResizeObserverRef.ts
```diff
@@ -20,26 +19,35 @@ const config: ResizeObserverOptions = {
...
 function useResizeObserverRef(
-  callback: ResizeObserverCallback,
+  callback: ResizeObserverCallback | undefined,
   options: ResizeObserverOptions = config
 ): [CallbackRef] {
+  const { box } = options;
   const [node, setNode] = useState<HTMLElementOrNull>(null);
-  const freshCallback = useFreshTick(callback);
+  const callbackRef = useRef(callback);
+  callbackRef.current = callback;
```
imbhargav5, line 28: This should be in a useEffect since it is a side effect. 😄  It's not a big deal, but it makes sense semantically to keep side effects out of the render phase. 🙏
```suggestion
 useEffect(() => {
      callbackRef.current = callback;
 }) 
```

superLipbalm: You are right. In principle, side effects should not be in the render phase, but in useEffect.
Thanks!

packages/rooks/src/hooks/useResizeObserverRef.ts
superLipbalm: Oh, that's right! `handleResizeObserver` is registered as a callback, so set a node without a callback then set a callback, the callback may not fire the node is not being observed! Thank you!

imbhargav5 (approved):

PR #1574, opened 2023-02-23
fix(useDebouncedValue): types of useDebouncedValue will only add null to the union if initializeWithNull was passed

### LINKED ISSUE #1573: types for useDebouncedValue
I've spotted that the types for `useDebouncedValue` are not ideal, as they alway return a union with `null`, even if we don't add the `initializeWithNull` option. Consider:

```
const [debouncedValue] = useDebouncedValue('string', 200)
```

`debouncedValue` will be of type `string | null`, even though it should be `string`

[TypeScript playground reproduction](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwK4GcCmARNAjCSB2AxmgCYBqAhgDZJoC+cAZlBCHAOQsQDWK7AsACghhCPhTwA2iRx4ipSjTQBdOAF44qTLILFy1WgAp2EqMHwBzdgBo4AJgAMDgJRCA9G7hevAPQD8QA)

the `| null` should only be added if we explicitly provide option `initializeWithNull: true`:

```
const [debouncedValue] = useDebouncedValue('string', 200, { initializeWithNull: true })
```

I think this can be solved with conditional return types and another generic. Happy to contribute this :)

TkDodo:
fixes #1573

### DIFF
--- a/packages/rooks/src/__tests__/useDebouncedValue.spec.ts
+++ b/packages/rooks/src/__tests__/useDebouncedValue.spec.ts
@@ -2,6 +2,16 @@ import type { RenderResult } from "@testing-library/react-hooks";
 import { act, renderHook } from "@testing-library/react-hooks";
 import { useDebouncedValue } from "@/hooks/useDebouncedValue";
 
+type Equal<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y
+  ? 1
+  : 2
+  ? true
+  : false;
+
+type Expect<T extends true> = T;
+
+const doNotExecute = (_func: () => void) => true;
+
 describe("useDebouncedValue", () => {
   beforeEach(() => {
     jest.useFakeTimers("modern");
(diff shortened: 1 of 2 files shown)


TkDodo:
I haven't found any other type tests so I just added the helpers I usually work with

packages/rooks/src/__tests__/useDebouncedValue.spec.ts
```diff
@@ -100,4 +110,42 @@ describe("useDebouncedValue", () => {
     });
     expect(result.current[0]).toBe(mockValue);
   });
+
+  describe("types", () => {
+    it("should return a union with null when initializeWithNull is true", () => {
+      doNotExecute(() => {
```
imbhargav5, line 116: Holy! I love this pattern of just testing types without running code. Very nice!

packages/rooks/src/hooks/useDebouncedValue.ts
```diff
@@ -4,17 +4,23 @@ import { useDebounce } from "./useDebounce";
...
 };
 
-type UseDebouncedValueReturnType<T> = [
-  debouncedValue: T | null,
-  immediatelyUpdateDebouncedValue: Dispatch<SetStateAction<T | null>>
+type UseDebouncedValueReturnType<
+  TValue = unknown,
+  TInitializeWithNull extends boolean = false
+> = [
+  debouncedValue: TInitializeWithNull extends true ? TValue | null : TValue,
```
imbhargav5, line 20: Very nice!

imbhargav5:
Love this @TkDodo ! Thank you very much!

PR #1714, opened 2023-04-17
useLocalstorageState: Ensure the identity of set/remove don't change when used

nathggns:
In https://github.com/imbhargav5/rooks/commit/adaaad1725977d539f5bda8c1fc8ee8358bd6a49#diff-b0add5bafe9b472c2eb73a766b1fde5062317a10e944f82725c1d782ccc7d6bf the possibility of using a function to generate the next state for `useLocalstorageState` was added. Unfortunately, this was done by adding the current value as a dependency of set, which meant its identity changed every time it is used. This is not the behaviour of `useState` and is not ideal – and most importantly its not necessary. We can just read the value from a ref instead.

This PR does that, and adds extra tests to ensure the identity of set (and remove) doesn't change when used.

### DIFF
--- a/.changeset/stale-poets-walk.md
+++ b/.changeset/stale-poets-walk.md
@@ -1,0 +1,5 @@
+---
+"rooks": patch
+---
+
+useLocalstorageState: ensure stable identity for set/remove functions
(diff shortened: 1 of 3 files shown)


imbhargav5:
Thanks @nathggns this is super useful! I appreciate it 🤩

PR #1726, opened 2023-06-16
Search and highlight word in the DOM

Amin-Partovi:
This PR introduces a new hook called useHighlight. The useHighlight hook accepts two arguments: keyword (the word we want to search for) and isReady (a boolean argument indicating whether the search should be performed). It returns a ref that should be assigned to the container element we want to search within.
I am ready to receive your feedback and address any comments you may have on my PR. Your time and consideration are greatly appreciated.


packages/rooks/src/hooks/useHighlight.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * useHighlight
+ * @description search for a keyword in the DOM and highlight them
+ * @see {@link https://rooks.vercel.app/docs/useHighlight}
+ */
+import { useEffect, useRef } from "react";
+
+function useHighlight({
```
imbhargav5, line 8: Super clever. I have a small preference for argument structure. Can we do something such as this. 

```tsx
useHighlight(keyword, {
   when: true
})
```

That way it will match some of the other hooks in our library. What do you think?

packages/rooks/src/hooks/useHighlight.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  useEffect(() => {
+    if (keyword && isReady && ref.current) {
+      const contentElement = ref.current;
+      if (contentElement) {
+        const text = contentElement.innerHTML;
+        const regex = new RegExp(keyword, "g");
+        const newText = text.replace(
+          regex,
+          `<mark class="highlighted">${keyword}</mark>`
```
imbhargav5, line 25: I was wondering if we can somehow get the DOM from another option and default the DOM to `<mark class="highlighted">{}</mark>`

```tsx
useHighlight(keyword, {
    render:(match) => `<mark class="highlighted">${keyword}</mark>`
})
```

imbhargav5, line 25: One thing I would like to add in the tests is that it should not highlight if it is already highlighted. For eg:
✅
`something keyword` => `something <mark class="highlighted">keyword</mark>` 
❌
`something <mark class="highlighted">keyword</mark>` => `something <mark class="highlighted"><mark class="highlighted">keyword</mark></mark>`

packages/rooks/src/hooks/useHighlight.ts
Amin-Partovi: Yes, I modify it for the consistency

packages/rooks/src/hooks/useHighlight.ts
Amin-Partovi: I wonder if I have understood your point correctly. I have added a render argument for cases where we want to include optional markup for the keywords.

Amin-Partovi:
@imbhargav5 
Hi Bhargav,

I hope you're doing well. I have addressed the comments you provided, and I was wondering if you could find some time to review my PR. I'm open to fixing any issues you mentioned and making any necessary adjustments.

Thank you for your time and consideration

PR #1952, opened 2026-06-23
types: narrow notification data option

imbhargav5:
## Summary
- Narrow `useNotification` option `data` from `any` to `unknown`

## Verification
- `pnpm vitest run src/__tests__/useNotification.spec.tsx` from `packages/rooks`


imbhargav5:
Closing as superseded: main now uses the platform NotificationOptions type, so this older standalone data-type change is no longer applicable.

PR #1956, opened 2026-06-23
types: type legacy color scheme listeners

imbhargav5:
## Summary
- replace usePreferredColorScheme legacy MediaQueryList @ts-ignore comments with a narrow local type
- keep the older-browser addListener/removeListener fallback behavior unchanged

## Tests
- pnpm exec vitest run src/__tests__/usePreferredColorScheme.spec.tsx
- pnpm run typecheck


imbhargav5:
@mergifyio queue

PR #1962, opened 2026-06-23
test: type useInViewRef observer mock

imbhargav5:
## Summary
- add a typed mock IntersectionObserver shape in the useInViewRef test
- replace loose any callback/options types in that test setup

## Verification
- pnpm --dir packages/rooks exec vitest run src/__tests__/useInViewRef.spec.tsx
- pnpm --dir packages/rooks run typecheck


imbhargav5:
@mergify queue

PR #1965, opened 2026-06-24
Improve Web Locks query return type

imbhargav5:
## Summary
- tighten useWebLocksApi query return type to LockManagerSnapshot

## Verification
- pnpm --filter rooks exec vitest run src/__tests__/useWebLocksApi.spec.tsx


imbhargav5:
@mergify queue

PR #1967, opened 2026-06-24
chore: remove selectable list debug log

imbhargav5:
## Summary\n- remove a stray console.log from useSelectableList's missing-value toggle path\n- assert the missing-value warning path does not emit debug logs\n\n## Verification\n- pnpm --filter rooks exec vitest run src/__tests__/useSelectableList.spec.tsx\n- pnpm --filter rooks typecheck\n


imbhargav5:
@mergifyio queue

PR #1972, opened 2026-06-24
docs: clarify useKeys when option

imbhargav5:
## Summary
- clarify the `useKeys` `when` option comment
- keep the change comment-only and scoped to one hook

## Verification
- pnpm --filter rooks exec vitest run src/__tests__/useKeys.spec.tsx


imbhargav5:
@mergifyio queue

PR #1974, opened 2026-06-24
types: fix session storage return alias typo

imbhargav5:
## Summary\n- rename the internal session storage hook return type alias to fix a typo\n\n## Verification\n- pnpm --filter rooks typecheck


imbhargav5:
@mergify queue

PR #1978, opened 2026-06-24
chore: type web locks query result

imbhargav5:
## Summary
- Type the Web Locks query helper as `LockManagerSnapshot`
- Let TypeScript infer the lock record type when filtering held and pending locks

## Verification
- `pnpm --filter rooks exec vitest run src/__tests__/useWebLocksApi.spec.tsx`


imbhargav5:
Closing as a duplicate of an earlier Codex PR covering the same change; keeping the oldest implementation in this series.

PR #1979, opened 2026-06-24
chore: type web locks query result

imbhargav5:
## Summary
- type the Web Locks query return value as LockManagerSnapshot
- type lock snapshot filters with LockInfo instead of any

## Verification
- pnpm --filter rooks exec vitest run src/__tests__/useWebLocksApi.spec.tsx
- git diff --check


imbhargav5:
Closing as duplicate of #1978 from the prior maintainer run; that PR already contains the same verified Web Locks type cleanup.

PR #1983, opened 2026-06-24
types: allow broadcast channel error callbacks

imbhargav5:
## Summary
- widen useBroadcastChannel onError callback type to accept Error as well as Event
- remove now-unneeded casts when forwarding caught errors

## Verification
- pnpm --filter rooks exec vitest run src/__tests__/useBroadcastChannel.spec.tsx


imbhargav5:
@mergifyio queue

PR #1988, opened 2026-06-25
chore: type web locks query results

imbhargav5:
## Summary
- Narrow the Web Locks query return type from any to LockManagerSnapshot
- Type held and pending lock filter callbacks with LockInfo

## Tests
- pnpm --dir packages/rooks exec vitest run src/__tests__/useWebLocksApi.spec.tsx
- pnpm --dir packages/rooks test


imbhargav5:
@mergifyio queue

PR #1998, opened 2026-06-25
Tighten broadcast channel default payload type

imbhargav5:
## Summary
- default useBroadcastChannel's generic payload type to unknown instead of any
- keep runtime behavior unchanged while improving the untyped callback surface

## Verification
- pnpm -C packages/rooks exec vitest run src/__tests__/useBroadcastChannel.spec.tsx
- pnpm -C packages/rooks typecheck


imbhargav5:
@mergify queue

PR #1999, opened 2026-06-25
test: type fullscreen mock element

imbhargav5:
## Summary
- replace the fullscreen test mock's any cast with a local mock element type
- remove a stray empty comment block from the same test

## Verification
- pnpm --dir packages/rooks exec vitest run src/__tests__/useFullscreen.spec.tsx


imbhargav5:
@mergify queue

imbhargav5:
@mergify queue

PR #2001, opened 2026-06-25
test: type useSpring raf mock callback

imbhargav5:
## Summary
- replace the `any` callback in the `useSpring` raf mock with an explicit timestamp callback type

## Verification
- pnpm --dir packages/rooks exec vitest run src/__tests__/useSpring.spec.ts
- pnpm --dir packages/rooks run typecheck


imbhargav5:
@mergify queue

imbhargav5:
@mergify queue

PR #2002, opened 2026-06-25
test: remove clipboard ts-ignore directives

imbhargav5:
## Summary
- replace repeated @ts-ignore comments in the useClipboard tests with a typed deleteClipboard helper
- keep the unsupported Clipboard API test behavior unchanged

## Verification
- pnpm --dir packages/rooks exec vitest run src/__tests__/useClipboard.spec.tsx


imbhargav5:
@mergify queue

PR #2003, opened 2026-06-25
fix session storage lazy initializer

imbhargav5:
## Summary
- call function initializers in useSessionstorageState when sessionStorage has no value
- add a focused regression test matching local storage lazy initializer behavior

## Verification
- pnpm --filter rooks exec vitest run src/__tests__/useSessionstorageState.spec.tsx


imbhargav5:
@mergifyio queue

PR #2004, opened 2026-06-25
test: tighten clipboard unsupported api suppressions

imbhargav5:
## Summary
- replace useClipboard test `@ts-ignore` comments with `@ts-expect-error` for unsupported Clipboard API scenarios
- document that the delete operations intentionally simulate browsers without Clipboard API support

## Tests
- `pnpm --dir packages/rooks exec vitest run src/__tests__/useClipboard.spec.tsx`
- `pnpm --dir packages/rooks run typecheck`


imbhargav5:
@mergifyio queue

PR #2021, opened 2026-06-26
test: type checkbox input change event

imbhargav5:
## Summary
- add an explicit ChangeEvent type import in the useCheckboxInputState test
- format the touched test file with Prettier

## Verification
- ./node_modules/.bin/vitest run src/__tests__/useCheckboxInputState.spec.tsx
- ./node_modules/.bin/tsc -p ./tsconfig.json --noEmit
- git diff --check


imbhargav5:
@mergify queue

PR #2033, opened 2026-06-27
test: fix window event listener test names

imbhargav5:
## Summary\n- update stale useWindowEventListener test names so they match the asserted behavior\n\n## Verification\n- cd packages/rooks && ../../packages/rooks/node_modules/.bin/vitest run src/__tests__/useWindowEventListener.spec.tsx


imbhargav5:
@mergifyio queue

PR #2050, opened 2026-06-28
fix(rooks): use safe own-property check in useMapState

imbhargav5:
## Summary
- replace the `hasOwnProperty` instance call in `useMapState.removeAll` with `Object.prototype.hasOwnProperty.call`
- remove the lint suppression tied to that check

## Verification
- `node_modules/.bin/vitest run src/__tests__/useMapState.spec.tsx`
- `node_modules/.bin/eslint src/hooks/useMapState.ts`


imbhargav5:
Closing as superseded: main now uses Object.keys for removeAll, which is already safe and simpler than this own-property check.

PR #2061, opened 2026-06-29
test: fix stale useOnWindowScroll spec description

imbhargav5:
### Summary\n- Update the stale scroll hook test description so it matches the event being exercised.\n\n### Verification\n- ./node_modules/.bin/vitest run src/__tests__/useOnWindowScroll.spec.ts\n


imbhargav5:
@mergify queue

PR #2068, opened 2026-06-30
test: remove ts-expect-error from useRafState spec

imbhargav5:
Small maintenance cleanup in the useRafState SSR test: replace a suppressed type error with a typed optional global override.

Verification:
- ./node_modules/.bin/vitest run src/__tests__/useRafState.spec.tsx


imbhargav5:
@mergify queue

PR #2071, opened 2026-06-30
chore(test): remove stray clipboard spec comment

imbhargav5:
## Summary\n- remove a stray empty JSDoc block from the clipboard test file\n\n## Verification\n- corepack pnpm@10.6.4 --dir packages/rooks exec vitest run src/__tests__/useClipboard.spec.tsx


imbhargav5:
@mergify queue

PR #2074, opened 2026-06-30
test: fix useKeys smoke assertion

imbhargav5:
Fix the shared smoke test so the useKeys block asserts the actual useKeys export instead of useKey. Verified with `vitest run src/__tests__/shared.spec.ts`.


imbhargav5:
@mergify queue

PR #2094, opened 2026-07-26
Remove unsafe user agent cache assertions

imbhargav5:
## Issue

The asynchronous cache handlers closed over the mutable `entry` binding, whose type includes `undefined`, and used four non-null assertions to suppress that possibility. This made the callbacks depend on an unchecked assumption even though the newly created cache entry was already available.

## Fix

Create a concrete `CacheEntry`, store it in the cache, assign it to the outer binding for the current render, and have the promise handlers update that concrete entry directly.

## Risk

Low. The same cache-entry object is created, stored, returned, and updated at the same points; the public API and runtime behavior are unchanged.

## Verification

- `pnpm --filter rooks exec vitest run src/__tests__/useSuspenseNavigatorUserAgentData.spec.tsx` (9 tests passed)
- `pnpm --filter rooks exec eslint src/hooks/useSuspenseNavigatorUserAgentData.ts`
- `git diff --check origin/main`

### DIFF
--- a/packages/rooks/src/hooks/useSuspenseNavigatorUserAgentData.ts
+++ b/packages/rooks/src/hooks/useSuspenseNavigatorUserAgentData.ts
@@ -134,22 +134,23 @@ function useSuspenseNavigatorUserAgentData(
     // Create new cache entry with promise
     const promise = navigator.userAgentData.getHighEntropyValues(hints);
     
-    entry = {
+    const cacheEntry: CacheEntry = {
       promise,
       status: 'pending'
     };
     
-    cache.set(cacheKey, entry);
+    cache.set(cacheKey, cacheEntry);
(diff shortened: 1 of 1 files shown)


imbhargav5 (commented):
PASS — no actionable findings.

- Scope is atomic: one commit and one file, replacing four non-null assertions with a typed local `CacheEntry`.
- Runtime behavior is preserved: the same object is inserted into the cache, assigned to the render-local binding, and updated by the fulfillment/rejection callbacks.
- Verification passed on head `62b5652`: package typecheck, focused test suite (9/9), focused ESLint, and `git diff --check`.
- GitHub's Node CI pull-request workflow and Vercel status are green.

Submitted as a non-approving comment because the authenticated reviewer is the PR author.

